PR #19, opened 2020-05-04
Generate Midgard API types

veado:
using [`open-api`](https://openapi-generator.tech/) and its [`typescript-rxjs` generator](https://openapi-generator.tech/docs/generators/typescript-rxjs)


src/types/generated/midgard/models/Coin.ts
```diff
@@ -0,0 +1,29 @@
...
+ * NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech).
+ * https://openapi-generator.tech
+ * Do not edit the class manually.
+ */
+
+/**
+ * @export
+ * @interface Coin
+ */
+export interface Coin {
```
robertdavid010, line 18: We should maybe use a more asset agnostic convention? Suggestions: 'Token' or 'Asset'(?)

veado: This code is generated by `open-api` and depends on definitions of Midgard endpoints. We can't change it.

Issue #41, opened 2020-05-11
ADD: Add Wallet (No Wallet)


thorchain-admin:
# Add wallet

The settings page is a canvas for a new wallet if none added:

* TESTNET | MAINNET tabs
*  IMPORT | CREATE Buttons
https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020?node-id=595%3A63

### Light Mode
<img width="932" alt="image" src="https://user-images.githubusercontent.com/41812729/81544895-6fad0980-93bb-11ea-8a5c-891d44d61edf.png">

### Dark Mode
<img width="935" alt="image" src="https://user-images.githubusercontent.com/41812729/81544999-9a975d80-93bb-11ea-9bcf-2fdb0cbc73d5.png">

PR #53, opened 2020-05-12
Issue 41/wallet settings view

robertdavid010:
Adding wallet settings views, types, and routes. Related to #41


src/views/wallet/WalletSettingsScreen.tsx
```diff
@@ -0,0 +1,162 @@
...
+  }, [])
+
+  const downloadLink: string = useMemo(() => {
+    const keystore: string = window.localStorage.getItem('binance') || ''
+    return 'data:text/plain;charset=utf-8,' + encodeURIComponent(keystore)
+  }, [])
+
+  const fileName = (): string => userAccount.address.concat('-keystore.txt')
+
+  const lockWallet = useCallback(async () => {
```
veado, line 57: [nitpick] This callback might be a placeholder for upcoming things, right? It does not need to by `async` for now. Or just put a comment to describe, it needs it be `async`.

src/views/wallet/WalletSettingsScreen.tsx
```diff
@@ -0,0 +1,162 @@
...
+    console.log('locking wallet...')
+  }, [])
+  const removeWallet = useCallback(async () => {
```
veado, line 60: [nitpick] Same here: Does it need to be `async` or not?

src/views/wallet/WalletSettingsScreen.tsx
```diff
@@ -0,0 +1,162 @@
...
+    console.log('removing wallet...')
+  }, [])
+
+  return (
+    <Row>
+      <Title level={3}>Settings</Title>
+      <Col span={24}>
+        <Row gutter={[16, 16]}>
+          <Col md={{ span: 12 }} lg={{ span: 12 }} xl={{ span: 12 }}>
```
veado, line 69: [nitpick] If `md={{ span: 12 }}` and `md ≥768px`, there is no need to define `lg` + `xl` with same values I guess... See https://ant.design/components/grid/#Col

src/views/wallet/WalletSettingsScreen.tsx
```diff
@@ -0,0 +1,162 @@
...
+                </Col>
+                <Col md={{ span: 24 }} lg={{ span: 12 }}>
+                  <Text strong>Chain ID:</Text>
+                  <Paragraph>{client.chainId}</Paragraph>
+                </Col>
+              </Row>
+            </Card>
+          </Col>
+
+          <Col sm={{ span: 24 }} md={{ span: 12 }} lg={{ span: 12 }} xl={{ span: 12 }}>
```
veado, line 126: [nitpick] Same as above: `md` should work (no need to use same values for `lg`, `xl`)

src/views/wallet/WalletSettingsScreen.tsx
robertdavid010: Good point. Not necessary for placeholders, since new wallet methods will be used, and not for now.

src/views/wallet/WalletSettingsScreen.tsx
robertdavid010: yeah left over from previous alterations. good find.

PR #63, opened 2020-05-15
Create wallet mnemonic confirmation

robertdavid010:
This is the new wallet mnemonic confirmation component and UI. It challenges the user to select randomized words from supplied mnemonic. Full interaction through all interface elements as well as custom error handling for mnemonic confirmation phrase.


src/components/wallet/NewMnemonicConfirm.tsx
```diff
@@ -0,0 +1,214 @@
...
+  const [shuffledWordsList, setShuffledWordsList] = useState<WordType[]>([])
+  const [loadingMsg] = useState<string>('')
+  const [mnemonicError, setMnemonicError] = useState<string>('')
+  const [initialized, setInitialized] = useState<boolean>(false)
+
+  function init() {
+    const words = mnemonic.split(' ')
+    if (words && !initialized) {
+      const res = words.map((e: string) => {
+        const uniqueId = Math.random().toString(36).substring(2) + Date.now().toString(36)
```
veado, line 26: This function to get a unique id looks unsafe, in a worse case it would return a same id. If you really want to have a unique id use a library such as [uniqid]( https://www.npmjs.com/package/uniqid) for that. Besides that, please extract any helper function from component into a `helpers/xxx` module and add tests. That's not necessary by using external libraries such as `uniqid`.

src/components/wallet/NewMnemonicConfirm.tsx
```diff
@@ -0,0 +1,214 @@
...
+        return { text: e, _id: uniqueId }
+      })
+      setWordsList(res)
+      setShuffledWordsList(shuffledWords(res))
+      setInitialized(true)
+    }
+  }
+
+  function shuffledWords(words: WordType[]) {
```
veado, line 35: Please extract `shuffledWords` into a `helpers/xxx` module and add some tests. Can be done w/ another PR.

src/components/wallet/NewMnemonicConfirm.tsx
```diff
@@ -0,0 +1,214 @@
...
+      for (let i = newArr.length - 1; i > 0; i--) {
+        const rand = Math.floor(Math.random() * (i + 1))
+        ;[newArr[i], newArr[rand]] = [newArr[rand], newArr[i]]
+      }
+      return newArr
+    }
+    return words.length > 0 ? shuffler(words) : []
+  }
+
+  function sortedSelectedWords(): WordType[] {
```
veado, line 47: Please extract `sortedSelectedWords` into a `helpers/xxx` module and add some tests. Can be done w/ another PR.

src/components/wallet/NewMnemonicConfirm.tsx
```diff
@@ -0,0 +1,214 @@
...
+        comparison = -1
+      }
+      return comparison
+    }
+    return wordsList.filter((e: WordType) => e.selected === true).sort(compare)
+  }
+
+  init()
+
+  const isSelected = (id: string) => {
```
veado, line 64: Please extract `isSelected` from component to export it and to add tests for it. It needs to be `isSelected = (id: string, wordsList)` then. Can be done w/ another PR.

src/components/wallet/NewMnemonicConfirm.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+    if (checkwords) {
+      // The submitted phrase as a string for passing to wallet methods
+      const repeatPhrase = wordsList
+        .filter((e: WordType) => e.selected === true)
+        .map((f: WordType) => {
+          return f.text
+        })
+        .join(' ')
+      console.log(repeatPhrase)
```
veado, line 148: Instead of logging `repeatPhrase` you could extract this snippet from component and add just a test for it. Please do it w/ another PR.

src/components/wallet/NewMnemonicConfirm.tsx
robertdavid010: ah ok good point. I did check the math, but your point to use a standard lib is appropriate.

src/components/wallet/NewMnemonicConfirm.tsx
robertdavid010: in new issue.

PR #80, opened 2020-05-19
Add label component & storybook

skyorion427:


src/components/uielements/label/label.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+        <Label size="large" color="dark">
+          This is the LABEL!
+        </Label>
+        <Label size="large" color="light">
+          This is the LABEL!
+        </Label>
+      </div>
+    )
+  })
+  .add('properties', () => {
```
veado, line 157: Since you have provided a story by using properties via knobs - which is great -, why do we still need stories for `small`, `normal`, `big`? I would suggest to remove other stories to have `properties` story only. Just keep it simple!

src/components/uielements/label/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  size: string
+  spacing: string
+}
+
+export type FontSettings = {
+  [key: string]: FontSetting
+}
+
+export type Colors = {
+  [key: string]: string
```
veado, line 15: `[key in Color]: string`

src/components/uielements/label/types.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export type SizeOptions = {
+  [key: string]: Size
```
veado, line 19: `[key in Size]: Size`

src/components/uielements/label/types.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export type WeightOptions = {
+  [key: string]: string
+}
+
+export type ColorOptions = {
+  [key: string]: Color
```
veado, line 27: `[key in Color]: Color`

src/components/uielements/label/types.ts
```diff
@@ -0,0 +1,28 @@
...
+}
```
veado, line 28: BTW: If you need `XXXOptions` for stories only, move these into `label.stories`

src/components/uielements/label/label.stories.tsx
skyorion427: Okay, I removed other stories

src/components/uielements/label/types.ts
skyorion427: Fixed

src/components/uielements/label/types.ts
skyorion427: Fixed

src/components/uielements/label/types.ts
skyorion427: Fixed

src/components/uielements/label/types.ts
skyorion427: Fixed

src/components/uielements/label/label.style.ts
```diff
@@ -1,8 +1,17 @@
...
+
+type Props = {
+  size?: Size
+  color?: Color
+  weight?: string
+  className?: string
+  children?: React.ReactNode
+  loading?: boolean
+  onClick?: () => void
+}
```
veado, line 14: Define `Props` for `label.style` only, which are really used for styles. For example:

```
type Props = {
  size?: Size
  color?: Color
  weight?: string
  usePointer?: boolean
}
```

skyorion427: Fixed!

PR #113, opened 2020-05-26
Issue 104/add mobile header

nomadnic:
Pending:

- Set tab from drawer (if necessary)
- Language dropdown placement when opened


src/components/header/Header.style.tsx
```diff
@@ -84,3 +92,60 @@ export const TabLink = styled(Link)`
...
+  align-items: center;
+  transition: none;
+  height: 60px;
+  display: flex;
+  text-transform: uppercase;
+  font-family: 'MainFontSemiBold';
+  font-size: 18px;
+  color: ${palette('text', 1)};
+`
+export const HeaderDrawerItemNoBorder = styled(Row)`
```
veado, line 135: You might not need to create another Component just to hide a border. Have you tried to use `:last-child` in `HeaderDrawerItem` to do the same thing?

src/components/header/Header.tsx
```diff
@@ -96,32 +95,92 @@ const Header: React.FC<Props> = (_): JSX.Element => {
     [items, activeKey, headerHeight]
   )
 
-  const color = useMemo(() => palette('text', 0)({ theme }), [theme])
+  const links = items.map(({ label, key, path, icon: Icon }) => (
```
veado, line 98: [suggestion] How about to use `useMemo` here (similar to `tabs` function)?

src/components/header/Header.tsx
```diff
@@ -96,32 +95,92 @@ const Header: React.FC<Props> = (_): JSX.Element => {
...
-          </Row>
-        </Col>
-        <Col>
-          <Row align="middle">
-            <ThemeIcon onClick={clickSwitchThemeHandler} style={{ color, ...iconStyle }} />
-            <SettingsIcon onClick={clickSettingsHandler} style={{ color, ...iconStyle }} />
-            <LockIcon onClick={clickLockHandler} style={{ ...iconStyle }} />
+      <>
+        <Row justify="space-between" align="middle" style={{ height: headerHeight }}>
+          <Col xs={0} sm={0} md={0} lg={24}>
```
veado, line 114: [nitpick] Just `xs={0} lg={24}` should be fine (not need to set other breakpoints w/ same values).

src/components/header/Header.tsx
```diff
@@ -96,32 +95,92 @@ const Header: React.FC<Props> = (_): JSX.Element => {
...
+                <Row align="middle">
+                  <HeaderTheme />
+                  <HeaderSettings />
+                  <HeaderLock />
+                  <HeaderLang />
+                </Row>
+              </Col>
+            </Row>
+          </Col>
+          <Col lg={0} xl={0}>
```
veado, line 137: Just setting `lg... ` would be fine. Even better than setting `span` to `0`: Use[ `useBreakpoint`](https://ant.design/components/grid/#components-grid-demo-useBreakpoint) hook to avoid to render this column. Because it would never goes into (virtual) DOM then. 

Check out the `Footer` to see an example how to use `useBreakpoint`.

src/components/header/HeaderLock.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { useThemeContext } from '../../contexts/ThemeContext'
+import { HeaderLockWrapper } from './HeaderLock.style'
+import { ReactComponent as SettingsIcon } from '../../assets/svg/icon-lock.svg'
+import { MobileWrapper } from './Header.style'
+
+type Props = {
+  onPress?: () => void
+}
+
+const HeaderLock: React.FC<Props> = (_: Props): JSX.Element => {
```
veado, line 13: `_` is just a placeholder for an used variable and to avoid an eslint error (see https://github.com/thorchain/asgardex-electron/blob/master/.eslintrc#L39).` Better way something like this:

```
const HeaderLock: React.FC<Props> = (props: Props): JSX.Element => {
// Destructure props and provide default value for `onPress` - see next comment
const { onPress = () => {} } = props
...
}
```

src/components/header/HeaderLock.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { theme$ } = useThemeContext()
+  const theme = useObservableState(theme$)
+
+  const color = useMemo(() => palette('text', 0)({ theme }), [theme])
+  const iconStyle = { fontSize: '1.5em', marginLeft: '20px' }
+
+  const clickHandler = useCallback(() => {
+    // has to be implemented
+    _.onPress && _.onPress()
```
veado, line 22: [nitpick] Set default value for `onPress` to avoid this check (see previous comment)

src/components/header/HeaderNetStatus.tsx
```diff
@@ -90,17 +91,64 @@ const HeaderNetStatus: React.FC<Props> = (_: Props): JSX.Element => {
     )
   }, [menuItems, theme])
 
+  type RootItemProps = { children: React.ReactNode }
+
+  const menuMobile = menuItems.map((item, i) => {
+    const { label, key, url } = item
+    const color = palette('text', 0)({ theme })
+    const RootElement = (props: RootItemProps) =>
+      i === menuItems.length - 1 ? (
```
veado, line 100: It might not needed if you just use `HeaderDrawerItem` (see my comment in `HeaderStyle`)

src/components/header/HeaderNetStatus.tsx
```diff
@@ -90,17 +91,64 @@ const HeaderNetStatus: React.FC<Props> = (_: Props): JSX.Element => {
...
+          {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+          <a className="ant-dropdown-link" onClick={(e) => e.preventDefault()}>
+            <Row justify="space-between" align="middle">
+              <ConnectionStatus color={onlineStatusColor} />
+              <DownIcon />
+            </Row>
+          </a>
+        </Dropdown>
+      </Col>
+      <Col lg={0} xl={0} style={{ width: '100%' }}>
```
veado, line 149: [nitpick] `xl={0}` is not needed. As mentioned before I would prefer to use `useBreakpoints` hook to not to render this `Col`.

src/components/header/HeaderSettings.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { HeaderSettingsWrapper } from './HeaderSettings.style'
+import * as walletRoutes from '../../routes/wallet'
+import { ReactComponent as SettingsIcon } from '../../assets/svg/icon-settings.svg'
+import { MobileWrapper } from './Header.style'
+
+type Props = {
+  onPress?: () => void
+}
+
+const HeaderSettings: React.FC<Props> = (_: Props): JSX.Element => {
```
veado, line 15: As mentioned before: `_` should be used as a placeholder.

src/components/header/HeaderSettings.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const { theme$ } = useThemeContext()
+  const theme = useObservableState(theme$)
+
+  const color = useMemo(() => palette('text', 0)({ theme }), [theme])
+  const iconStyle = { fontSize: '1.5em', marginLeft: '20px' }
+
+  const history = useHistory()
+  const clickSettingsHandler = useCallback(() => {
+    history.push(walletRoutes.settings.path())
+    _.onPress && _.onPress()
```
veado, line 25: Provide default value for `onPress` to avoid this check.

src/components/header/HeaderTheme.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useThemeContext } from '../../contexts/ThemeContext'
+import { HeaderThemeWrapper } from './HeaderTheme.style'
+import { ReactComponent as ThemeIcon } from '../../assets/svg/icon-theme-switch.svg'
+import { MobileWrapper } from './Header.style'
+
+type Props = {
+  onPress?: () => void
+}
+
+const HeaderTheme: React.FC<Props> = (_: Props): JSX.Element => {
```
veado, line 13: `_` === placeholder

src/components/header/HeaderTheme.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const { toggleTheme, theme$ } = useThemeContext()
+  const theme = useObservableState(theme$)
+
+  const color = useMemo(() => palette('text', 0)({ theme }), [theme])
+  const iconStyle = { fontSize: '1.5em', marginLeft: '20px' }
+
+  const clickSwitchThemeHandler = () => {
+    toggleTheme()
+    // _.onPress && _.onPress()
```
veado, line 22: Remove this line ^

src/components/header/Header.tsx
```diff
@@ -111,8 +116,8 @@ const Header: React.FC<Props> = (_): JSX.Element => {
     <HeaderContainer>
       <>
         <Row justify="space-between" align="middle" style={{ height: headerHeight }}>
-          <Col xs={0} sm={0} md={0} lg={24}>
-            <Row justify="space-between" align="middle" style={{ height: headerHeight }}>
+          {(screens.lg || screens.xl || screens.xxl) && (
```
veado, line 119: [suggestion] Instead of checking all breakpoints, how about something like this:

```
const isDesktopView = screens.lg
...
{isDesktopView && (...)}
```

src/components/header/Header.tsx
```diff
@@ -132,10 +137,10 @@ const Header: React.FC<Props> = (_): JSX.Element => {
                   <HeaderLang />
                 </Row>
               </Col>
-            </Row>
-          </Col>
-          <Col lg={0} xl={0}>
-            <Row justify="space-between" align="middle" style={{ height: headerHeight, width: '100vw' }}>
+            </>
+          )}
+          {!screens.lg && !screens.xl && !screens.xxl && (
```
veado, line 142: Based on previous comment ^ you could try this:
```
{!isDesktopView && (...)}
```

src/components/header/Header.tsx
```diff
@@ -150,36 +155,36 @@ const Header: React.FC<Props> = (_): JSX.Element => {
...
-            <HeaderLock onPress={() => closeMenu()} />
-          </HeaderDrawerItem>
-          <HeaderDrawerItem>
-            <HeaderSettings onPress={() => closeMenu()} />
-          </HeaderDrawerItem>
-          <HeaderDrawerItem>
-            <HeaderLang />
-          </HeaderDrawerItem>
-          <HeaderDrawerItemNoBorder>
+        {!screens.lg && !screens.xl && !screens.xxl && (
```
veado, line 161: {!isDesktopView && (...)}

src/components/header/HeaderLock.tsx
```diff
@@ -4,27 +4,28 @@ import { useObservableState } from 'observable-hooks'
...
     // has to be implemented
-    _.onPress && _.onPress()
-  }, [_])
+    onPress()
+  }, [onPress])
 
   return (
     <HeaderLockWrapper onClick={() => clickHandler()}>
-      <MobileWrapper>LOCK WALLET</MobileWrapper>
+      {!screens.lg && !screens.xl && !screens.xxl && 'LOCK WALLET'}
```
veado, line 28: Just  `{!screens.lg && 'LOCK WALLET'}` should be good.

src/components/header/HeaderTheme.tsx
```diff
@@ -4,26 +4,31 @@ import { useObservableState } from 'observable-hooks'
...
+    onPress()
   }
+
   return (
     <HeaderThemeWrapper onClick={() => clickSwitchThemeHandler()}>
-      <MobileWrapper>{palette('background', 0)({ theme }) === '#fff' ? 'DAY MODE' : 'NIGHT MODE'}</MobileWrapper>
+      {!screens.lg &&
+        !screens.xl &&
+        !screens.xxl &&
+        (palette('background', 0)({ theme }) === '#fff' ? 'DAY MODE' : 'NIGHT MODE')}
```
veado, line 31: `{!screens.lg && (palette...}` should be fine...

PR #117, opened 2020-05-29
Add Password Confirmation Modal

FrozenIce0617:
- Add Password Confirmation Modal


src/components/modal/privateModal/privateModal.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { useCallback } from 'react'
+
+import { LockOutlined } from '@ant-design/icons'
+import { Form } from 'antd'
+
+import Input from '../../uielements/input'
+import { StyledModal } from './privateModal.style'
+
+interface Props {
```
veado, line 9: Pls change `interface` to `type`. You can't know it, but we have decided to use `type` (not `interface`) anymore. See https://gitlab.com/thorchain/bepswap/bepswap-web-ui/-/issues/325

PR #121, opened 2020-06-01
Add PoolStatus component

nomadnic:


src/components/uielements/poolStep/PoolStatus.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Row } from 'antd'
+import BigNumber from 'bignumber.js'
+import { useIntl } from 'react-intl'
+
+import Label from '../label'
+import Trend from '../trend'
+import { PoolStatusWrapper } from './PoolStatus.style'
+
+type Props = {
+  amount: BigNumber
```
veado, line 12: For amounts we are using `BaseAmount` or `AssetAmount` from `asgardex-util` https://gitlab.com/thorchain/asgardex-common/asgardex-util/-/blob/master/src/types.ts#L19-20 I would suggest to use `AssetAmount` here. Checkout [`asset.test.ts`](https://gitlab.com/thorchain/asgardex-common/asgardex-util/-/blob/master/src/asset.test.ts) to see examples about how to use one of these ^.

src/components/uielements/poolStep/PoolStatus.tsx
```diff
@@ -0,0 +1,38 @@
...
+  asset: string
+  label: string
+  target: string
+  trend: BigNumber
+}
+
+const PoolStatus: React.FC<Props> = (props: Props): JSX.Element => {
+  const { amount, asset, label, target, trend } = props
+  const intl = useIntl()
+  const amountVal = intl.formatNumber(amount.toNumber(), { currency: 'USD', style: 'currency' })
```
veado, line 22: If we are using `AssetAmount` we can use a format helper [`formatAssetAmount`](https://gitlab.com/thorchain/asgardex-common/asgardex-util/-/blob/master/src/asset.ts#L70) or [`formatAssetAmountCurrency`](https://gitlab.com/thorchain/asgardex-common/asgardex-util/-/blob/master/src/asset.ts#L75-79). It depends on what you want to show here.

PR #123, opened 2020-06-01
Provide all pool data

### LINKED ISSUE #109: Populate `PoolView` w/ pool detail data
- Extend `MidgardService` to fetch data
- Provide data to `PoolView` via `MidgardContext`

veado:
- [x] `MidgardService`: 
     - Loading "queue" for all needed of pool data 
     - Prepare and store all data into `PoolsState` 
     - Provide all needed types
- [x] Simplify `MidgardContext` to provide `MidgardService` only
- [x] Add example into `PlaygroundView` to load + display all pool data. 
- [x] Migrate `src/services/midgard/utils.test.ts` + `src/services/midgard/utils.test.ts` from BEPSwap - that's basically just "copy / paste".
- [x] Add `toPoolDetailsMap` incl. test to  `src/services/midgard/utils.test.ts`

Closes #109


src/views/playground/PlaygroundView.tsx
```diff
@@ -21,9 +22,9 @@ const PlaygroundView: React.FC<Props> = (_): JSX.Element => {
 
   const tickers = useObservableState(miniTickers$, [])
 
-  useEffect(() => {
-    console.log('tickers', tickers[0]?.s)
-  }, [tickers])
+  // useEffect(() => {
+  //   console.log('tickers', tickers[0]?.s)
+  // }, [tickers])
```
nomadnic, line 27: Asking for clarification purposes - why is this included but commented out?

src/services/binance/service.ts
```diff
@@ -71,7 +71,7 @@ export const subscribeTransfers = (address: string) => {
       (e) => (e as WS.TransferEvent).data !== undefined
     )
     .pipe(
-      tap((msg) => console.log('transfer msg', msg)),
+      // tap((msg) => console.log('transfer msg', msg)),
```
nomadnic, line 74: What purpose does this serve?

src/services/binance/service.ts
```diff
@@ -106,7 +106,7 @@ export const miniTickers$ = ws$
     (e) => (e as WS.MiniTickersEvent).data !== undefined
   )
   .pipe(
-    tap((ev) => console.log('mini tickers msg', ev)),
+    // tap((ev) => console.log('mini tickers msg', ev)),
```
nomadnic, line 109: Same as previous.

src/services/binance/service.ts
veado: It's for logging the value of this websocket message. By using [`tap`](https://rxjs.dev/api/operators/tap), you can do something with the value of the Observable and it will still return the same Obervable so you won't lost it for the next function (it's `map` in the next line). 

Here are some nice examples to explain `tap`: https://www.learnrxjs.io/learn-rxjs/operators/utility/do#examples

src/services/binance/service.ts
veado: See https://github.com/thorchain/asgardex-electron/pull/123#discussion_r434388839

src/views/playground/PlaygroundView.tsx
veado: Good catch - it can be removed.

PR #128, opened 2020-06-02
Add TxTimer component

FrozenIce0617:
- Add TxTimer component


src/assets/timer/refund.svg
```diff
@@ -0,0 +1,51 @@
...
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
+<g>
+</g>
```
nomadnic, line 50: These are probably not needed 😆

src/components/uielements/txTimer/TxTimer.tsx
```diff
@@ -0,0 +1,121 @@
...
+          className={CircularProgressbarStyle}
+          value={value}
+          text={`${totalDurationString}s`}
+          strokeWidth={7}
+          counterClockwise
+          styles={buildStyles({
+            textColor: '#23DCC8',
+            textSize: '14px',
+            pathColor: '#23DCC8',
+            trailColor: '#F3F4F4',
```
nomadnic, line 112: Should be able to reference these colors from the palette
https://gitlab.com/thorchain/asgardex-common/asgardex-theme/-/blob/master/src/palette.ts

PR #140, opened 2020-06-03
Update routes to new structure

### LINKED ISSUE #137: Update routes to new structure
Before
```
├── Swap
├── Stake
├── Wallet
```
Now

```
├── Pools
│   ├── Swap
│   └── Stake
├── Wallet
```

^ Needed by #87

veado:
- `swap` + `stake` are nested routes of `pools` now
- Update tests for routes
- Add empty `PoolsView`

Closes #137


src/routes/swap.ts
```diff
@@ -12,7 +13,7 @@ export const swap: Route<SwapRouteParams> = {
   template: `${base.template}/:source-:target`,
   path: ({ source, target }) => {
     if (source && target) {
-      return `/swap/${source.toLowerCase()}-${target.toLowerCase()}`
+      return `${base.template}/${source.toLowerCase()}-${target.toLowerCase()}`
     } else {
```
FrozenIce0617, line 17: We don't need `else` here, I think.

veado: Good catch, thx!

Issue #101, opened 2020-05-24
Add PoolShare Card component


skyorion427:
![image](https://user-images.githubusercontent.com/20286132/82743062-af68df00-9d33-11ea-8ef3-d1e28931f5ac.png)

PR #143, opened 2020-06-03
Add PoolShare component

nomadnic:


nomadnic:
Closes #101

nomadnic:
This component is a subset of the PoolStake component  in bepswap. 

I rewrote the implementation/styling to be more aligned with asgardex. This commit is essentially the difference between what was in bepswap and my implementation: https://github.com/thorchain/asgardex-electron/pull/143/commits/20bb34a44bce22be2b4082e4f4c320f30474e61c

@veado 
I'm not 100% sure on the types used for the different values (..Share/..Price) in PoolShare.tsx. Maybe you can comment here.

src/components/uielements/label/label.style.ts
```diff
@@ -48,6 +49,7 @@ const colors: Colors = {
 
 export const LabelWrapper = styled.div`
   padding: 10px 0;
+  text-align: ${(props: Props) => props.align};
```
FrozenIce0617, line 52: `props.align` can be `undefined`.
I think it would be great if you can add default value here.

nomadnic: The default is defined in the src/components/uielements/label/label.tsx component. This should be good enough right? See https://github.com/thorchain/asgardex-electron/blob/20bb34a44bce22be2b4082e4f4c320f30474e61c/src/components/uielements/label/label.tsx#L20

veado: @FrozenIce0617 is right (good catch btw). A default value in `label.tsx` won't save you, as long as `align?: TextAlignment` is defined as an optional value.

FrozenIce0617: If the `align` has a default value, we can define `align` as a required props. :)

src/components/uielements/label/label.style.ts
```diff
@@ -55,4 +56,5 @@ export const LabelWrapper = styled.div<Props>`
   letter-spacing: ${({ size = 'normal' }) => fontSettings[size].spacing};
   color: ${({ color }) => colors[color || 'normal']};
   cursor: ${({ onClick }) => onClick && 'pointer'};
+  text-align: ${(props: Props) => props.align || 'left'};
```
veado, line 59: [hint] 
```ts
text-align: ${({ align = 'left' }) => align};
```

Issue #148, opened 2020-06-05
Save phrase encrypted to disk


veado:
by using `thorchain-crypto` (not published yet)

![image](https://user-images.githubusercontent.com/61792675/83878838-065fb280-a73d-11ea-9799-d5e7e01b94e7.png)

PR #154, opened 2020-06-05
Lock / unlock wallet

### LINKED ISSUE #149: Lock / unlock wallet

veado:
- [x] Update `WalletContenxt` + `wallet/service` to handle `locked` state
- [x] Add new wallets routes `wallet/locked` + `wallet/imports` (incl. tests)
- [x] Refactor `WalletView` to handle routes properly depending on `locked` + `phrase` state
- [x] Add `LockView` (not styled yet)
- [x] Update `WalletManage` component to `lock` + `remove` wallet (This component does still need more work on its logic - it will be another PR) 
- [x] Fix `HeaderLock` to represent correct icons (desktop + mobile views are different)
- [x] Update header to toggling `lock` | `unlock` state
- [x]  Add + update icons `unlock` / `lock`
- [x] Fix storybook (missing `WalletProvider`)

**Important Note:** Currently `phrase` is stored as a plain, **not** decrypted text. That's just a temporary approach and will be fixed in #148 .

Closes #149


src/components/wallet/WalletManage.tsx
```diff
@@ -43,6 +44,9 @@ const WalletManage: React.FC = (): JSX.Element => {
   const [network, setNetwork] = useState<string | null>()
   const [chainId, setChainId] = useState<string | null>()
   const [address, setAddress] = useState<string | null>('')
+
+  const { lock, phrase } = useWalletContext()
+
   async function setData() {
```
FrozenIce0617, line 50: How about using arrow function?
```const setData = async () => {```

veado: Good point! Since that's legacy code (you can't know it) and `WalletManage` has to be refactored in either way, we will do it then.

PR #163, opened 2020-06-08
Add unit tests for PoolView

FrozenIce0617:
- Add Unit Test for `getPoolViewData`
- Add Unit Test for `getPoolData`


src/views/pools/util.test.ts
```diff
@@ -0,0 +1,178 @@
...
+          target: 'FSN'
+        },
+        poolPrice: 'RUNE 2.000',
+        depth: 'RUNE 0.00',
+        volume: 'RUNE 0.00',
+        transaction: 'RUNE 0.00',
+        slip: '0',
+        trade: '0',
+        raw: {
+          depth: new BigNumber(200000),
```
veado, line 110: [hint] You can use a shorthand function `bn()` here (and anywhere else) instead of `new BigNumber`.

src/views/pools/util.test.ts
```diff
@@ -0,0 +1,178 @@
...
+        slip: '50100000',
+        trade: '1',
+        raw: {
+          depth: new BigNumber(199999799),
+          volume: new BigNumber(0),
+          transaction: new BigNumber(16193),
+          slip: new BigNumber(50100000),
+          trade: new BigNumber(1),
+          poolPrice: new BigNumber(0)
+        }
```
veado, line 156: When I see this test, I think `getPoolData` function should provide `raw` data only, but not formatted value of `raw` data. All formatted values can be easily created from `raw` data in a component by using format utility functions we have in `asgardex-util`. What do you think @FrozenIce0617 ?

veado: ^ Moved into https://github.com/thorchain/asgardex-electron/issues/168

PR #165, opened 2020-06-08
Add AssetData component

FrozenIce0617:
- Add AssetData component

### DIFF
--- a/src/components/uielements/assets/assetData/AssetData.style.tsx
+++ b/src/components/uielements/assets/assetData/AssetData.style.tsx
@@ -1,9 +1,6 @@
 import styled from 'styled-components'
 import { palette } from 'styled-theme'
 
-type AssetDataWrapperProps = {
-  target?: boolean
-}
 export const AssetDataWrapper = styled.div`
   display: flex;
   align-items: center;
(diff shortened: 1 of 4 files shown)


src/components/uielements/assets/assetData/AssetData.style.tsx
```diff
@@ -0,0 +1,41 @@
+import styled from 'styled-components'
+import { palette } from 'styled-theme'
+
+type AssetDataWrapperProps = {
+  target?: boolean
```
veado, line 5: Do we still need `target` ? We never set it in `AssetData ` component.

src/components/uielements/assets/assetData/AssetData.tsx
```diff
@@ -0,0 +1,31 @@
...
+  assetValue?: number
+  price?: number
+  priceValue?: string
+  priceUnit: string
+  size?: CoinSize
+  className?: string
+}
+
+const AssetData: React.FC<Props> = (props: Props): JSX.Element => {
+  const { asset, priceValue, priceUnit, size = 'big', className = '', ...otherProps } = props
```
veado, line 18: Try to avoid to use `...otherProps`, it's just unsafe. _I know we did it in BEPSwap all over the place, but we should change now..._ I think there is a `lint` rule for this, which we should use to avoid it.

src/components/uielements/assets/assetData/AssetData.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  return (
+    <AssetDataWrapper className={`assetData-wrapper ${className}`} {...otherProps}>
```
veado, line 21: Same here: Try to avoid `...otherProps`.

PR #174, opened 2020-06-12
Simplify calculations for pools

### LINKED ISSUE #172: Calculate pool values using `calc` helpers of `asgardex-utils`
Currently all calculation of pool data still based on BEPSwap's implementation. Since we have [`calc` helpers of `asgardex-utils`](https://gitlab.com/thorchain/asgardex-common/asgardex-util/-/tree/master/src/calc), we can simplify things.

veado:
- [x] Use `calc` helper functions from `asgardex-util`
- [x] Format values in `PoolsOverview` (but not in `getPoolData` to avoid
duplication of data
- [x] Update + simplify misc. tests
- [x] Simplify `PoolsState` (remove `PriceDataIndex` + `AssetDetailMap`)
- [x] Remove `getPriceIndex` helper
- [x] Add `toPoolData` helper (incl. test)
- [x] Quick fix: `round` property of button did not work
- [x] Use latest `asgardex-util`

Closes #172

### DIFF
--- a/src/views/pools/utils.ts
+++ b/src/views/pools/utils.ts
@@ -3,10 +3,9 @@ import {
   PoolData,
   assetAmount,
   getValueOfAsset1InAsset2,
-  // getValueOfRuneInAsset,
   baseAmount,
-  baseToAsset,
-  getValueOfRuneInAsset
+  getValueOfRuneInAsset,
+  assetToBase
 } from '@thorchain/asgardex-util'
 
 import { RUNE_TICKER } from '../../const'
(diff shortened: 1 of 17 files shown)


src/views/pools/utils.ts
```diff
@@ -1,58 +1,54 @@
-import { baseAmount, formatBaseAsTokenAmount } from '@thorchain/asgardex-token'
-import { validBNOrZero, bn } from '@thorchain/asgardex-util'
+import {
+  bnOrZero,
+  PoolData,
+  assetAmount,
+  getValueOfAsset1InAsset2,
+  // getValueOfRuneInAsset,
```
nomadnic, line 6: Commented-out import

veado: Good catch, thx!

PR #181, opened 2020-06-15
Add `HeaderCurrency` component

veado:
- [x] Refactor styles of `HeaderLang` to share common styles

# Desktop view
Screenshot from 2020-06-15 19-06-48

# Mobile view
Screenshot from 2020-06-15 19-07-32


src/services/midgard/types.ts
```diff
@@ -27,3 +28,10 @@ export type PoolsState = {
 export type PoolsStateRD = RD.RemoteData<Error, PoolsState>
 
 export type NetworkInfoRD = RD.RemoteData<Error, NetworkInfo>
+
+export type PoolCurrency = {
+  symbol: string
+  name: string
+  poolData: PoolData
+}
+export type PoolCurrencies = PoolCurrency
```
veado, line 37: You can ignore types of `PoolCurrency` + `PoolCurrencies`. These are needed to deal with real data for `HeaderCurrency`, but that will be another PR.

src/components/header/HeaderCurrency.stories..tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
```
nomadnic, line 1: File name should be HeaderCurrency.stories.tsx

PR #184, opened 2020-06-16
Add AssetSelect component

FrozenIce0617:
- Add AssetSelect component

(Port `components/uielements/token/tokenSelect` component from `BEPSWAP`)

- [x] Remove legacy attributes (like `data-test`)
- [x] Remove rest props (`...otherProps`)
- [x] Use Nick's helper function (`useClickOutside` function)
- [x] Use `useRef` instead of `Ref` helper

### DIFF
--- a/src/components/uielements/assets/assetSelect/AssetSelectMenu.tsx
+++ b/src/components/uielements/assets/assetSelect/AssetSelectMenu.tsx
@@ -54,7 +54,7 @@ const AssetSelectMenu: React.FC<Props> = (props: Props): JSX.Element => {
   const disableItemFilterHandler = useCallback(
     (item: AssetPair) => {
       const tokenName = getTickerFormat(item.asset)?.toLowerCase()
-      return searchDisable.indexOf(tokenName ?? '') > -1
+      return searchDisable.indexOf(tokenName || '') > -1
     },
     [searchDisable]
   )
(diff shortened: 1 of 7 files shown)


src/components/uielements/assets/assetSelect/AssetSelectMenu.tsx
```diff
@@ -0,0 +1,74 @@
...
+      const node = <CoinData asset={tokenName || ''} price={price} priceUnit={unit} />
+      return { key, node }
+    },
+    [priceIndex, unit]
+  )
+
+  const disableItemFilterHandler = useCallback(
+    (item: AssetPair) => {
+      const tokenName = getTickerFormat(item.asset)?.toLowerCase()
+      return searchDisable.indexOf(tokenName ?? '') > -1
```
veado, line 57: [nitpick] `||` instead of `??` here.

It will be generated following code
```
tokenName !== null && tokenName !== void 0 ? tokenName : ''
```
see [TS Playground Example](https://www.typescriptlang.org/play/#code/DYUwLgBGD2DWIDsByBDAtiAXBAzmATgJYIDmA3AFADG0CeUI9AvBABQCUETAfBAN4UIQiPnABXfAihxEqDBAD8CiAHIVFAL4UgA)

PR #193, opened 2020-06-18
Resolve "Fix: Pool data are loaded twice at start"

### LINKED ISSUE #191: Fix: Pool data are loaded twice at start
Screenshot from 2020-06-18 10-23-16

veado:
Closes #191

### DIFF
--- a/src/helpers/stateHelper.ts
+++ b/src/helpers/stateHelper.ts
@@ -1,6 +1,6 @@
 import { Observable, BehaviorSubject } from 'rxjs'
 
-type ObservableState<T> = {
+export type ObservableState<T> = {
   get$: Observable<T>
   get: () => T
   set: (value: T) => void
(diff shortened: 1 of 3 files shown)


src/helpers/stateHelper.ts
```diff
@@ -36,3 +36,21 @@ export const observableState = <T>(initial: T): ObservableState<T> => {
...
+  trigger: () => void
+}
+
+/**
+ * Helper to create a stream, which can trigger changes of it
+ *
+ * It might be handy to trigger other, depending streams in a queue to do something
+ */
+export const triggerStream = (): TriggerStream => {
+  const subject$$ = new BehaviorSubject('')
```
nomadnic, line 51: I'm not yet familiar with RxJS. Should empty string be the correct value here?

veado: Yep, that's just to provide any data here. It could be anything in this `BehaviorSubject` object as long as it has the same type called in `next()`. This data is hidden for the outside world and it's just used to "trigger" a "change" of the subject (`BehaviorSubject`).

PR #195, opened 2020-06-18
Save / load keystore

### LINKED ISSUE #148: Save phrase encrypted to disk

veado:
Depends on https://gitlab.com/thorchain/asgardex-common/asgardex-crypto/-/merge_requests/1

- [x] Save / load keystore
- [x] Update handling of lock / unlock wallet 
- [x] Fix wallet route handling 
- [x] Quick fix: Error "Uncaught (in promise) Invalid BIP39 phrase passed to Binance Client" (will be implemented by https://github.com/thorchain/asgardex-electron/issues/197)

## Info

After importing the phrase and created a keystore successfully, we save the keystore as a file on disk on following location (called `appData` in Electron):

> appData Per-user application data directory, which by default points to:
> %APPDATA%/ASGARDEX/keystore.json on Windows
> $XDG_CONFIG_HOME/ASGARDEX/keystore.json or ~/.config/ASGARDEX/keystore.json on Linux
> ~/Library/Application Support/ASGARDEX/keystore.json on macOS
> ^ see https://www.electronjs.org/docs/api/app#appgetpathname

The keystore file will include the phrase encrypted (this feature is provided by asgardex-crypto). After importing the phrase, the decrypted phrase is stored in memory as long as the wallet has not been locked. By locking the wallet, the phrase will be removed from memory. To unlock the wallet, a password is needed to load and decrypt the phrase from stored keystore again. ~~If the app has been closed, the keystore file will be removed from disk (that will be another PR #199).~~

Closes #148


thorchain-admin:
"If the app has been closed, the keystore file will be removed from disk"

This is not the intent, keep the keystore always on disk. Only remove it if they specifically "REMOVE WALLET" which is the red button on the settings.

veado:
> "If the app has been closed, the keystore file will be removed from disk"
> 
> This is not the intent, keep the keystore always on disk. Only remove it if they specifically "REMOVE WALLET" which is the red button on the settings.

Ok, #199 has been closed.

PR #210, opened 2020-06-23
Setup BinanceClient + Load balances

veado:
- [x] Update `BinanceContext` + `binance/service` to support initialization of `BinanceClient` using imported phrase
- [x] Provide handling for (re-)loading assets from Binance
- [x] Update `UserAssetScreen` to populate asset table with data
- [x] Fix: `Uncaught (in promise) invalid password`
- [x] Fix: `BinanceClient` should be initialized as soon as phrase has been imported. With this an error handling of `BinanceClient.init()` can be already handled in `ImportPhrase` view (but not later by reqesting data in `UserAssetView`.
- [x] Fix: Reload of balances
- [x] ~~Fix: `clientState`~~


src/services/binance/service.ts
```diff
@@ -1,18 +1,27 @@
...
 const BINANCE_TESTNET_WS_URI = envOrDefault(
   process.env.REACT_APP_BINANCE_TESTNET_WS_URI,
   'wss://testnet-dex.binance.org/api/ws'
 )
 
-export const BINANCE_MAINET_WS_URI = envOrDefault(
-  process.env.REACT_APP_BINANCE_MAINNET_WS_URI,
-  'wss://dex.binance.org/api/ws'
-)
+const BINANCE_MAINET_WS_URI = envOrDefault(process.env.REACT_APP_BINANCE_MAINNET_WS_URI, 'wss://dex.binance.org/api/ws')
```
nomadnic, line 24: Would it be worth storing these defaults in one place?

veado: Good catch: In this case we just need it in `services/binance/service.ts`

PR #211, opened 2020-06-23
Style ImportPhrase

### LINKED ISSUE #156: Style `ImportPhrase`
Layout: https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020?node-id=1306%3A1520
Source: `src/components/wallet/ImportPhrase.tsx`

nomadnic:
Close #156


nomadnic:
Some comments:

- The design has a full height background - I wasn't able to accomplish this without using flexblox and setting the app's main layout wrapper to display: 'flex'. This had too many side effects so I kept it out

- I have some inline styling on the form items - not sure if this is fine or if you'd rather have each form item wrapper in a styled-component?

- I introduced Input.Password to the Input component which requires the prop security='password'. See ImportPhrase.tsx for this

src/components/uielements/input/InputTextArea.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
```
veado, line 1: Same here: Why do we need this `InputTextArea` component? It just wrapped `InputTextAreaWrapper`. A better approach would be just to define `InputTextArea` in `Input.style` and any other component can import it if needed.

src/components/uielements/input/Input.tsx
```diff
@@ -2,22 +2,28 @@ import React from 'react'
 
 import { InputProps } from 'antd/lib/input'
 
-import { InputWrapper } from './Input.style'
-import { Color, InputType, Size } from './types'
+import { InputWrapper, InputPasswordWrapper } from './Input.style'
```
veado, line 5: Why do we need this `Input` component? It just wrapped `InputPasswordWrapper` / `InputWrapper` and both could be imported into any other component it if needed.

src/components/wallet/ImportPhrase.tsx
```diff
@@ -94,36 +93,33 @@ const ImportPhrase: React.FC = (): JSX.Element => {
...
+      <Form form={form} onFinish={submitForm} labelCol={{ span: 24 }} style={{ padding: 30, paddingTop: 15 }}>
         {/* TODO(@Veado): i18n */}
         <Spin spinning={importing} tip="Loading...">
           <Form.Item
             name="phrase"
             rules={[{ required: true, validator: phraseValidator }]}
             validateTrigger={['onSubmit', 'onChange']}>
             {/* TODO(@Veado): i18n */}
-            <Input.Password placeholder="Enter your phrase" size="large" />
+            <InputTextArea placeholder="ENTER PHRASE" rows={5} />
```
veado, line 104: Add any text values into our `i18n` sources. Please do the same for all other texts in this PR.

veado:
>     * The design has a full height background - I wasn't able to accomplish this without using flexblox and setting the app's main layout wrapper to display: 'flex'. This had too many side effects so I kept it out

That's fine!

>     * I have some inline styling on the form items - not sure if this is fine or if you'd rather have each form item wrapper in a styled-component?

That's ok, too.
> 
>     * I introduced Input.Password to the Input component which requires the prop security='password'. See ImportPhrase.tsx for this. I wasn't able to completely align the styling between the password and normal component. Some of the style props aren't applying for Input.Password

Ok.

src/components/uielements/input/Input.tsx
nomadnic: I guess that's true.. The Input component already existed in the project so I just extended on it.

src/components/uielements/input/InputTextArea.tsx
nomadnic: I initially did it that way. The problem is that the Input component extends InputProps and the Input.TextArea component extends TextAreaProps. It breaks when trying to spread InputProps over the Input.TextArea component. 

image

src/components/uielements/backLink/BackLink.tsx
```diff
@@ -1,25 +1,25 @@
...
 
 import { BackLinkWrapper } from './BackLink.style'
 
 type Props = {
-  route: string
   title?: string
 }
 
-const BackLink: React.FC<Props> = ({ route, title = 'Back', ...otherProps }): JSX.Element => {
+const BackLink: React.FC<Props> = ({ title = 'Back', ...otherProps }): JSX.Element => {
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: agree here. in general IMAO its better not to use spread at all in case you know the props type. its more explicit in render itself. for example
```ts
const Foo: React.FC<{click: () => void}> = ({click} ) => {
  // a lot of code here
  return <button onClick={click}>click me</button>
}
```
in this case i don't know that this click handler is just a proxied value from props and can suppose that this is some 'computed' value calculated somewhere inside of the component's function. And this point is also relevant during reviewing PRs when you dont see whole function's body and have to unwrap whole function's body to understand that it was just a value from props.

more expicit way to use props' values directly

```ts
const Foo: React.FC<{click: () => void}> = (props ) => {
  // a lot of code here
  return <button onClick={props.click}>click me</button>
}
```

src/components/uielements/input/InputTextArea.tsx
thatStrangeGuyThorchain: @veado im not sure that this is a better idea. in this case there is gonna be a 'monster' component after some time which should resolve what it should use with a lot of `if` conditions. i believe it's gonna be really hard maintainable. antd decided the same way and splitted it to separate components with providing static methods:
image

src/views/wallet/importsView/ImportsView.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type Tab = {
+  key: TabKey
+  label: string
+  content: React.ReactNode
+}
+
+const ImportsView: React.FC = (): JSX.Element => {
+  const items = useMemo(() => [{ key: TabKey.PHRASE, label: 'Phrase', content: <ImportPhrase /> }] as Tab[], [])
```
thatStrangeGuyThorchain, line 22: i18n for label?

thatStrangeGuyThorchain, line 22: its better not to typecast interfaces manually.
in case of typecast with `as` you allowed to do this thing
```ts
const items = useMemo(() => [{ label: 'Phrase', content: <ImportPhrase /> }] as Tab[], []) // hard to notice when you debugging anything that there is missed key value
// or even this
const items = useMemo(() => [{}] as Tab[], []) // more obvious but anyway
```
better to use explicit way without casting
```ts
const items: Tab[] = useMemo(() => [{ key: TabKey.PHRASE, label: 'Phrase', content: <ImportPhrase /> }], [])
```

veado: yes

thatStrangeGuyThorchain:
@veado check please. also im not sure about translations

src/components/wallet/ImportPhrase.tsx
```diff
@@ -101,14 +103,27 @@ const ImportPhrase: React.FC = (): JSX.Element => {
             rules={[{ required: true, validator: phraseValidator }]}
             validateTrigger={['onSubmit', 'onChange']}>
             {/* TODO(@Veado): i18n */}
-            <InputTextArea placeholder="ENTER PHRASE" rows={5} />
+            <InputTextArea
```
veado, line 106: ^ TODO can be removed

src/components/wallet/ImportPhrase.tsx
```diff
@@ -101,14 +103,27 @@ const ImportPhrase: React.FC = (): JSX.Element => {
...
+              rows={5}
+            />
           </Form.Item>
           <Form.Item
             name="password"
             rules={[{ required: true, validator: passwordValidator }]}
             validateTrigger={['onSubmit', 'onChange']}>
             {/* TODO(@Veado): i18n */}
-            <Input placeholder="PASSWORD" security="password" style={{ width: 280 }} />
+            <InputPasswordWrapper
```
veado, line 119: ^ Pls remove TODO

src/i18n/types.ts
```diff
@@ -27,6 +28,8 @@ export type WalletMessages = {
   'wallet.column.value': string
   'wallet.action.send': string
   'wallet.action.receive': string
+  'wallet.imports.phrase': string
+  'wallet.imports.enter.phrase': string
```
veado, line 33: [nitpick] `wallet.imports.enterphrase` (`enterphrase` instead of `enter.phrase`)

src/components/uielements/input/Input.tsx
```diff
@@ -2,28 +2,22 @@ import React from 'react'
...
   typevalue = 'normal',
-  security = 'normal',
   ...rest
 }): JSX.Element => {
-  return security === 'password' ? (
-    <InputPasswordWrapper color={color} sizevalue={sizevalue} typevalue={typevalue} {...rest} />
-  ) : (
-    <InputWrapper color={color} sizevalue={sizevalue} typevalue={typevalue} {...rest} />
-  )
+  return <InputWrapper color={color} sizevalue={sizevalue} typevalue={typevalue} {...rest} />
```
veado: I still think we can get rid of `Input.tsx` component by using a `styled(Input)` component only (not tested):

```
# In `Input.styles.tsx`

import * as A from 'antd/lib/input'

type CustomInputProps {
  sizevalue?: Size
  color?: Color
  typevalue?: InputType
}

export type InputProps = CustomInputProps & A.InputProps

export const Input = styled(A.Input)<InputProps>`
 //  ... add all styles here
`
```


```
// in any other component which needs the `Input` component

import * as Styled from './Input.styled'

<Styled.Input />

thatStrangeGuyThorchain: yes. we can. the only thing here is that here are default values for 3 props. this component is used in 15 files and everywhere we should set this props manually 'cause setting defaultProps for styled component does not work with TS now.

veado: Fixed with b6bbb8d135cffd61aab07a9560c66ee04c629085

src/components/uielements/input/Input.tsx
veado: Resolved with b6bbb8d135cffd61aab07a9560c66ee04c629085

PR #231, opened 2020-07-02
Feature/asset details

FrozenIce0617:
- [x] Style AssetDetails page
- [x] Update Coin & Button component (for size)
- [x] Add `i18n` for AssetDetails page
- [x] Update TransactionTable component

### DIFF
--- a/src/i18n/de/wallet.ts
+++ b/src/i18n/de/wallet.ts
@@ -8,7 +8,14 @@ const wallet: WalletMessages = {
   'wallet.column.name': 'Name - DE',
   'wallet.column.ticker': 'Ticker - DE',
   'wallet.column.balance': 'Balance - DE',
-  'wallet.column.value': 'Value - DE'
+  'wallet.column.value': 'Value - DE',
+  'wallet.action.send': 'Send - DE',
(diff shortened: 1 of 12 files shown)


src/i18n/de/wallet.ts
```diff
@@ -8,7 +8,14 @@ const wallet: WalletMessages = {
...
   'wallet.column.balance': 'Balance - DE',
-  'wallet.column.value': 'Value - DE'
+  'wallet.column.value': 'Value - DE',
+  'wallet.action.send': 'Send - DE',
+  'wallet.action.receive': 'Receive - DE',
+  'wallet.transaction.table.type': 'Type - DE',
+  'wallet.transaction.table.address': 'Address - DE',
+  'wallet.transaction.table.to': 'To - DE',
+  'wallet.transaction.table.amount': 'Amount - DE',
+  'wallet.transaction.table.coin': 'Coin - DE'
```
veado, line 18: Most of these translations are very "common" and might be used at other places as well, such as `Name`, `Amount`, `Value` etc. Put translations into `wallet` only, if we don't need these at other places. Or move these translations into `i18n/{lang}/common`. 

We don't need it now, but we should keep that in mind.

Issue #236, opened 2020-07-03
Optimize `render` functions in tables


veado:
to avoid re-render leaks.

veado:
Resolved by #240

PR #240, opened 2020-07-06
Optimize `render` functions in tables #236

thatStrangeGuyThorchain:
- [x]  optimize render functions

### DIFF
--- a/src/components/wallet/UserTransactionsTable.tsx
+++ b/src/components/wallet/UserTransactionsTable.tsx
@@ -14,6 +14,7 @@ type Props = {
 
 type Column = 'type' | 'address' | 'to' | 'amount' | 'coin'
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 const getColumnsRenderers = (address: string): Record<Column, (value: any, tx: UserTransactionType) => JSX.Element> => {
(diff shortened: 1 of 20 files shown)


src/components/wallet/UserTransactionsTable.tsx
```diff
@@ -12,63 +12,79 @@ type Props = {
   transactions: UserTransactionType[]
 }
 
-const TransactionsTable: React.FC<Props> = ({ transactions }): JSX.Element => {
-  const intl = useIntl()
+type Column = 'type' | 'address' | 'to' | 'amount' | 'coin'
 
-  const address = 'tbnb1vxutrxadm0utajduxfr6wd9kqfalv0dg2wnx5y'
+const getColumnsRenderers = (address: string): Record<Column, (value: any, tx: UserTransactionType) => JSX.Element> => {
```
veado, line 17: Try to avoid `any`.

thatStrangeGuyThorchain: @veado i totally agree with but here is no any chance to that. Internal antd's table typings  provide only this.
image
we also can not use `unknown` here 'casue in this case there will not be a possibility to type `coin` renderer

PR #245, opened 2020-07-08
[Issue 244] Asset-View: Re-position of "refresh" button

thatStrangeGuyThorchain:
- [x] wrapped ViewRoutes to common View
- [x] moved refresh button outside to the WalletView
- [x] added i18n for refresh button
- [x] added debounce time to triggerStream helper

### DIFF
--- a/src/i18n/de/common.ts
+++ b/src/i18n/de/common.ts
@@ -13,7 +13,7 @@ const common: CommonMessages = {
   'common.coin': 'Coin - DE',
   'common.password': 'Passwort',
   'common.memo': 'Memo - DE',
-  'common.refresh': 'AKTUALISIERUNG'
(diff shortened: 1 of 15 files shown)


src/i18n/de/common.ts
```diff
@@ -12,7 +12,8 @@ const common: CommonMessages = {
   'common.amount': 'Amount - DE',
   'common.coin': 'Coin - DE',
   'common.password': 'Passwort',
-  'common.memo': 'Memo - DE'
+  'common.memo': 'Memo - DE',
+  'common.refresh': 'AKTUALISIERUNG'
```
veado, line 16: Capitalize texts w/ CSS (via `text-transform`) in a component, but not in `i18n` sources.

PR #248, opened 2020-07-08
[issues/65] Extract shuffle words from mnemonic confir

### LINKED ISSUE #65: Extract shuffle words from mnemonic confirm
As mentioned:

https://github.com/thorchain/asgardex-electron/pull/63#discussion_r426452858

thatStrangeGuyThorchain:
- [x] moved shuffleArray to separate helper

Closes #65


src/components/wallet/NewMnemonicConfirm.tsx
```diff
@@ -5,6 +5,8 @@ import { Card, Col, Row, Button, Form } from 'antd'
 import { Store } from 'antd/lib/form/interface'
 import { v4 as uuidv4 } from 'uuid'
 
+import { shuffleArray } from '../../helpers/shuffleArrayHelper'
```
veado, line 8: Similar question[ to previous PR](https://github.com/thorchain/asgardex-electron/pull/249): How about to rename function + module names:
- `helpers/shuffleArrayHelper` -> `helpers/array` or `helpers/list`
- `shuffleArray` -> `shuffle`
?

src/helpers/shuffleArrayHelper.ts
```diff
@@ -0,0 +1,11 @@
+export const shuffleArray = <T extends unknown>(array: T[]) => {
+  const newArr = array.slice()
+
+  for (let i = newArr.length - 1; i > 0; i--) {
+    const rand = Math.floor(Math.random() * (i + 1))
```
veado, line 5: `Math.random()` is unsafe to get always a "real" `random` value. How about to use `[lodash.shuffle](https://www.npmjs.com/package/lodash.shuffle)` (based on [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher-Yates_shuffle) or maybe another package with less dependencies?

thatStrangeGuyThorchain: yeah. make sense. but i just moved this code outside.
and the second thing is that real randoms do not matter here. next line swaps 2 'whatever elements'. here we dont care about how much random is real

veado: `lodash.shuffle` or similar would save us to create this helper and its test here, too.

PR #249, opened 2020-07-08
[issue-67] extract isselected method/snippet from mnemonic confirm

thatStrangeGuyThorchain:
- [x] moved isSelected to separate factory helper

### DIFF
--- a/src/components/wallet/NewMnemonicConfirm.tsx
+++ b/src/components/wallet/NewMnemonicConfirm.tsx
@@ -5,6 +5,8 @@ import { Card, Col, Row, Button, Form } from 'antd'
 import { Store } from 'antd/lib/form/interface'
 import { v4 as uuidv4 } from 'uuid'
 
+import { isSelectedFactory } from '../../helpers/isSelectedHelper'
+
 type WordType = {
   text: string
   _id: string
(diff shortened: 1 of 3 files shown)


src/components/wallet/NewMnemonicConfirm.tsx
```diff
@@ -5,6 +5,8 @@ import { Card, Col, Row, Button, Form } from 'antd'
 import { Store } from 'antd/lib/form/interface'
 import { v4 as uuidv4 } from 'uuid'
 
+import { isSelectedFactory } from '../../helpers/isSelectedHelper'
```
veado, line 8: How about to name this helper just `isSelected` (w/o `factory`) and the module `helpers/arrayHelper` or `helpers/listHelper`?

src/components/wallet/NewMnemonicConfirm.tsx
```diff
@@ -65,13 +67,8 @@ const MnemonicConfirmScreen: React.FC<{ mnemonic: string; onConfirm: Function }>
...
   init()
 
-  const isSelected = useCallback(
-    (id: string) => {
-      const res: WordType | undefined = wordsList.find((e: WordType) => e._id === id)
-      return !!res?.selected
-    },
-    [wordsList]
-  )
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
veado, line 70: Why do we need this ^ here?

thatStrangeGuyThorchain: react's preset linter can not infer returned type from factory function return type and throws a warning that deps list is unknown

src/components/wallet/NewMnemonicConfirm.tsx
thatStrangeGuyThorchain: but.... this is a factory for isSelected functions😅

veado: Ah, I've missed that. `factory` is fine in this case ;)

PR #253, opened 2020-07-09
Hide "Refresh" button on wallet settings screen #251

### LINKED ISSUE #251: Hide "Refresh" button on wallet settings screen
Screenshot from 2020-07-09 14-51-33

thatStrangeGuyThorchain:
- [x] added styles for asset name at assetDetails
- [x] removed refresh button from settings screen

Fixes #251

### DIFF
--- a/src/views/wallet/WalletView.styles.tsx
+++ b/src/views/wallet/WalletView.styles.tsx
@@ -1,0 +1,10 @@
+import styled from 'styled-components'
+
+import Button from '../../components/uielements/button'
+
+export const RefreshButton = styled(Button)`
+  border: none;
+  background: none !important;
+  margin-bottom: 20px;
+  box-shadow: none;
+`
--- a/src/views/wallet/WalletView.tsx
+++ b/src/views/wallet/WalletView.tsx
@@ -6,7 +6,6 @@ import { useObservableState } from 'observable-hooks'
 import { useIntl } from 'react-intl'
 import { Switch, Route, Redirect } from 'react-router-dom'
 
-import Button from '../../components/uielements/button'
 import AssetsNav from '../../components/wallet/AssetsNav'
 import { useBinanceContext } from '../../contexts/BinanceContext'
 import { useWalletContext } from '../../contexts/WalletContext'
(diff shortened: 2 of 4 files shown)


src/views/wallet/WalletView.tsx
```diff
@@ -34,15 +34,23 @@ const WalletView: React.FC = (): JSX.Element => {
   // we have to add 'undefined'  as default value
   const keystore = useObservableState(keystoreService.keystore$, undefined)
 
+  const reloadButton = useMemo(
+    () => (
+      <div style={{ display: 'flex', justifyContent: 'flex-end' }}>
+        <Styled.RefreshButton typevalue={'outline'} onClick={reloadBalances}>
```
veado, line 40: [nitpick] We don't want an outline for the button anymore.

thatStrangeGuyThorchain: @veado i left it for the sake of saving another styles. background and border reset at the `RefresButton` styles

veado: Cool, thx!

src/views/wallet/WalletView.styles.tsx
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components'
+
+import Button from '../../components/uielements/button'
+
+export const RefreshButton = styled(Button)`
+  border: none;
+  background: none !important;
```
FrozenIce0617, line 7: `!important` is required?

src/views/wallet/WalletView.tsx
```diff
@@ -34,15 +34,23 @@ const WalletView: React.FC = (): JSX.Element => {
   // we have to add 'undefined'  as default value
   const keystore = useObservableState(keystoreService.keystore$, undefined)
 
+  const reloadButton = useMemo(
+    () => (
+      <div style={{ display: 'flex', justifyContent: 'flex-end' }}>
```
FrozenIce0617, line 39: We have to avoid inline styling since we're using `styled-components`.

src/views/wallet/WalletView.styles.tsx
thatStrangeGuyThorchain: unfortunately yes

src/views/wallet/WalletView.tsx
thatStrangeGuyThorchain: yep. you are right here. but i did not want to add another HOC here for just alignments

veado: @thatStrangeGuyThorchain You could use `Row` from `antd` here - no need for another styled component.

PR #254, opened 2020-07-09
Simplify usage of `Asset`

### LINKED ISSUE #252: Simplify usage of `Asset`
by using [latest changes of `asgardex-util`](https://gitlab.com/thorchain/asgardex-common/asgardex-util/-/merge_requests/11).

veado:
~~**Pls DON'T MERGE**: This PR depends [on latest change of `@thorchain/asgardex-util`](https://gitlab.com/thorchain/asgardex-common/asgardex-util/-/merge_requests/11/), which has to be published to npm before. _That's why CI is failing..._~~

Closes #252


src/components/uielements/assets/assetMenu/AssetMenu.tsx
```diff
@@ -25,34 +25,25 @@ const AssetMenu: React.FC<Props> = (props: Props): JSX.Element => {
   const { assetData, asset, priceIndex = {}, withSearch, searchDisable = [], onSelect = () => {} } = props
 
   const filteredData = useMemo(
-    () =>
-      assetData.filter((item) => {
-        const { ticker = '' } = item.asset
-        return ticker !== asset?.ticker
-      }),
+    () => assetData.filter((item) => item.asset.ticker !== asset?.ticker),
```
thatStrangeGuyThorchain, line 28: is still asset a nullable value

veado: You are right, it's a `NonNullable` asset now

thatStrangeGuyThorchain: so maybe we dont need `asset?` ?

veado: I think it would be more safety to pass an `Asset` we already have instead of a `string` into this component. A `string` could be anything else, but no guarantee to have a `ticker` of an `Asset`.

veado: Or did you mean the `?.` operator for optional `ticker`. Yes, it can be removed then.

PR #256, opened 2020-07-09
Fix style of refresh button

veado:
by providing a "transparent" button


src/views/wallet/WalletView.tsx
```diff
@@ -37,7 +37,7 @@ const WalletView: React.FC = (): JSX.Element => {
   const reloadButton = useMemo(
     () => (
       <div style={{ display: 'flex', justifyContent: 'flex-end' }}>
-        <Styled.RefreshButton typevalue={'outline'} onClick={reloadBalances}>
+        <Styled.RefreshButton typevalue="transparent" type="text" onClick={reloadBalances}>
```
veado, line 40: `type="text"  is still needed to avoid an animation of a border around while clicking the button. I have not found any other way to fix it.

Issue #250, opened 2020-07-09
Add `useRemoteImage` hook


veado:
as suggested by @thatStrangeGuyThorchain  in https://github.com/thorchain/asgardex-electron/pull/241#discussion_r451729563

PR #266, opened 2020-07-10
Add `useRemoteImage` hook #250

thatStrangeGuyThorchain:
- [x] added useRemote hook
- [x] added tests for useRemote hook

closes https://github.com/thorchain/asgardex-electron/issues/250

### DIFF
--- a/src/components/uielements/assets/assetIcon/AssetIcon.tsx
+++ b/src/components/uielements/assets/assetIcon/AssetIcon.tsx
@@ -1,10 +1,12 @@
-import React, { useState, useMemo, useCallback } from 'react'
+import React, { useMemo, useCallback } from 'react'
 
+import * as RD from '@devexperts/remote-data-ts'
 import { Asset, AssetTicker } from '@thorchain/asgardex-util'
 
 import bnbIcon from '../../../../assets/svg/coin-bnb.svg'
 import runeIcon from '../../../../assets/svg/rune-flash-icon.svg'
 import { getIntFromName, rainbowStop } from '../../../../helpers/colorHelpers'
+import { useRemoteImage } from '../../../../hooks/useRemoteImage'
 import * as Styled from './AssetIcon.style'
 import { Size } from './types'
 
@@ -15,7 +17,6 @@ interface Props extends React.HTMLAttributes<HTMLDivElement> {
 
 const AssetIcon: React.FC<Props> = (props: Props): JSX.Element => {
   const { asset, size = 'normal' } = props
-  const [error, setError] = useState(false)
 
   const imgUrl = useMemo(() => {
     const { symbol, ticker } = asset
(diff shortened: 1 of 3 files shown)


src/hooks/useRemoteImage.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    Object.defineProperty(global.Image.prototype, 'src', {
+      set(src: string) {
+        if (src === LOAD_FAILURE_SRC) {
+          this.onerror()
+        } else if (src === LOAD_SUCCESS_SRC) {
+          this.onload()
+        }
+      }
+    })
+  })
```
veado, line 26: As with other mocks or stubs, I would recommend to bring `Image` back to it's original "state" in `afterAll` or so.

thatStrangeGuyThorchain: @veado nice catch

src/hooks/useRemoteImage.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const useRemoteImage = (src: string) => {
+  const [remoteImage, setRemoteImage] = useState<RD.RemoteData<string, string>>(RD.pending)
+
+  useEffect(() => {
+    const image = new Image()
+    image.onload = () => setRemoteImage(RD.success(src))
+    image.onerror = () => setRemoteImage(RD.failure(src))
+    image.src = src
+  }, [src])
```
veado, line 13: What do you think about to clean-up everything like this?

```
  useEffect(() => {

    const onLoad = () => setRemoteImage(RD.success(src))
    const onError = () => setRemoteImage(RD.failure(src))

    const image = new Image()   
    image.addEventListener('load', onLoad)
    image.addEventListener('error', onError)
    image.src = src

   // clean-up callbacks
    return () => {
       image.removeEventListener('load', onLoad)
       image.removeEventListener('error', onError)
    }
  }, [src])

File: src/components/uielements/assets/assetIcon/AssetIcon.tsx [hunk 3]
```diff
@@ -63,12 +62,7 @@ const AssetIcon: React.FC<Props> = (props: Props): JSX.Element => {
     )
   }, [asset, size])
 
-  return (
-    <>
-      {!error && renderIcon}
-      {error && renderFallbackIcon}
-    </>
-  )
+  return RD.fold(() => <></>, renderFallbackIcon, renderFallbackIcon, renderIcon)(remoteIconImage)
```
veado, line 65: I would render a fallback icon on error only and for or `initial` and `loading` states `Styled.IconBG`

src/hooks/useRemoteImage.ts
thatStrangeGuyThorchain: there is no closures here and image object will cleaned from memory the same time as Unmount. with `image.onload` and `image.onerror` properties. so i believe it's not necessary to use listeners here

src/components/uielements/assets/assetIcon/AssetIcon.tsx
thatStrangeGuyThorchain: @veado there is no initial state by default behaviour of `useRemoteImage` (initial value is `pending`). for loading state here is a fallback renderer just 'cause i did not have a loading icon which should be placed. 
anyway i will place here whatever icon you say. just replace a few symbols.
just say exactly which ones ?:)

veado: For loading state you can use<IconWrapper><Styled.IconBG /><IconWrapper>, where [Styled.IconBG is just a "circle"](https://github.com/thorchain/asgardex-electron/blob/e6461d10c36cd2f68693931f1db5abd735ed5fd5/src/components/uielements/assets/assetIcon/AssetIcon.style.ts#L30-L37) with a size of an icon.

veado: Or we could use[ a custom `Spin` provided by `antd`](https://ant.design/components/spin/#components-spin-demo-custom-indicator) - it's up to you ;)

Peek 2020-07-11 19-15

Issue #270, opened 2020-07-11
Style Unlock View


veado:
Layout https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020?node-id=595%3A63

PR #276, opened 2020-07-13
Style Unlock View #270

thatStrangeGuyThorchain:
- [x] refactored App's layout. removed fixed positions from header and footer

- [x] fixed mobile menu layout
  
- [x] changed paddings for page content block
  
- [x] added hiding backlinks text on mobile small devices
  
- [x] styled unlock wallet view
  
- [x] added base translations to UnlockForm

closes https://github.com/thorchain/asgardex-electron/issues/270

### DIFF
--- a/src/components/wallet/UnlockForm.style.tsx
+++ b/src/components/wallet/UnlockForm.style.tsx
@@ -1,0 +1,90 @@
+import AF from 'antd/lib/form'
+import AT from 'antd/lib/typography/Text'
+import styled from 'styled-components'
+import { palette } from 'styled-theme'
+
+import { media } from '../../helpers/styleHelper'
+import BaseButton from '../uielements/button'
+
+export const Form = styled(AF)`
+  display: flex;
+  flex-direction: column;
+
+  ${media.lg`
+    height: 100%;
+  `}
+
+  .ant-form-item {
+    width: 100%;
+    margin: 0;
+  }
+`
+
+export const Header = styled('div')`
+  display: flex;
+  justify-content: center;
+  position: relative;
+
+  .ant-typography {
+    font-weight: 600;
+  }
+`
+
+export const Text = styled(AT)`
+  width: 100%;
+  text-align: center;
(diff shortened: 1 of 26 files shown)


veado:
## Use one breakpoint for mobile screens only.

I would suggest to use the breakpoint for "very small screens" only :arrow_down: 

## Very small screens:
- No need to remove space (margin-top??) from the row, which includes back button and  headline "Unlock you wallet". 
- Could you try to size the white bacḱground to 100% (similar to large screen)?

Screenshot from 2020-07-14 06-58-51

## Small screens:

- Please use same layout (100%) as we have in large screens == remove styles for this breakpoint.

Screenshot from 2020-07-14 06-58-35

veado:
## Fix styles of "Remove" button (dark theme)

- hover state: Change backround to grey (not white) 

Screenshot from 2020-07-14 07-07-21

src/components/wallet/UnlockForm.style.tsx
```diff
@@ -0,0 +1,90 @@
...
+    margin: 0;
+  }
+`
+
+export const Header = styled('div')`
+  display: flex;
+  justify-content: center;
+  position: relative;
+
+  .ant-typography {
```
veado, line 28: Is there another way to style fonts here? Accessing `antd`  style selectors such as `.ant-typography` are always unsafe. I know we already have that in different places, mostly by adapting styles from BEPSwap.

src/components/wallet/UnlockForm.style.tsx
```diff
@@ -0,0 +1,90 @@
...
+  align-items: center;
+  height: 100%;
+  padding: 45px 30px 35px 30px;
+  background-color: ${palette('background', 1)};
+
+  ${media.sm`
+    padding: 90px 60px 70px 60px;
+  `}
+
+  .ant-row {
```
veado, line 61: Same here: Is there another way to style this instead of using an `antd` class selector?

PR #282, opened 2020-07-14
Re-structure sources used by main and renderer processes

### LINKED ISSUE #275: Restructure sources used by `main` and `renderer` processes
### Before

```bash
├── electron # sources of main process
├── src      # sources of renderer process

```

### New
```bash
├──  src
      ├── main      # main process
      ├── renderer  # sources of renderer process
      └── shared    # shared sources process
```

veado:
A lot of changes here, but 99.9% is just moving content around to have sources in a new structure:

```bash
├──  src
      ├── main      # sources of main process
      ├── renderer  # sources of renderer process
      └── shared    # shared sources
```
Depends on #281

Closes #275


craco.config.js
```diff
@@ -1,3 +1,5 @@
+// const path = require('path')
```
thatStrangeGuyThorchain, line 1: maybe remove ?

tsconfig.main.json
```diff
@@ -8,8 +8,9 @@
     "strict": true,
     "module": "CommonJS",
     "moduleResolution": "node",
-    "outDir": "../public/",
+    "outDir": "./public/",
     "pretty": true
   },
-  "include": ["./", "../src/shared"]
+  "include": ["./src/main", "./src/shared"],
+  "exclude": ["src/remderer/"]
```
thatStrangeGuyThorchain, line 15: a typo at reMderer

craco.config.js
veado: Good catch!

tsconfig.main.json
veado: Another good catch :+1:

veado:
> have a question about renaming `electron` -> `main`. if i understood correctly main is still an electron app. maybe leave a semantic name ?

Folders `main` and `renderer` are named after Electrons [Main and Renderer Processes](https://www.electronjs.org/docs/tutorial/application-architecture#main-and-renderer-processes). We have already used these naming for packaging: [`electron-renderer` ](https://github.com/thorchain/asgardex-electron/blob/d953230a4079360181ede9e5190898520b1299b2/craco.config.js#L8) and [`electron-main`](https://github.com/thorchain/asgardex-electron/blob/2f1dc77c6b85d8e508d9c85b48932693372fd518/webpack.electron.js#L23). You will also find `ipcRenderer` or `mainRenderer` in Electron. I think it's still fine to use these names for our folders, which are all in the sense of Electron.

thatStrangeGuyThorchain:
> Folders main and renderer are named after Electrons Main and Renderer Processes. We have already used these naming for packaging: electron-renderer and electron-main. You will also find ipcRenderer or ipcMain in Electron (Note: xyzRenderer, xyzMain). I think it's still fine to use these names for our folders, which are all in the sense of Electron.

@veado thanks for explanation

Issue #273, opened 2020-07-13
Link Mnemonic view from ImportsView 


veado:
Currently Mnemonic View is not accessible, but it should be linked from "ImportsView" ([layout]( https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020?node-id=1306%3A6254) )

thatStrangeGuyThorchain:
@veado what is logic for Mnemonic ? is user available to type a string manully ?
if yes what is a scenario for user interaction ?

thorchain-admin:
@thatStrangeGuyThorchain 

Mnemonics are 12 or 24 word phrases that come from the BIP39 word list of 2048 English words (https://github.com/bitcoin/bips/blob/master/bip-0039/english.txt). 

The first 11 (or 22) words are entropy, the final 1 (2) words are a checksum, so they user MUST not pick their own words. 

they should use the BIP39 library to `generateMnemonic()` https://iancoleman.io/bip39/ which will return a valid phrase. 

However, they may have previously generated a valid phrase in another wallet and have funds on it. In which case they can add it and call `validateMnemonic(phrase)` in order to check that it is valid. 

----
The logic is:

Generate Phrase:
1) user clicks "new" and a list of 12 words appear. 
2) They then need to copy it down separately (secure notepad) then re-enter it
3) To make (2) easier, then can just select from a mix of words and demonstrate it is in correct order

Import Phrase:
1) User enters their 12 word phrase, it is simply checked

thorchain-admin:
Note: we are using our own library `asgardex-crypto` to handle everything, please review:

https://gitlab.com/thorchain/asgardex-common/asgardex-crypto#basic-usage

PR #285, opened 2020-07-16
Link Mnemonic view from ImportsView #273

thatStrangeGuyThorchain:
- [x] Currently Mnemonic View is not accessible, but it should be linked from "ImportsView"

closes https://github.com/thorchain/asgardex-electron/issues/273


src/renderer/views/wallet/NoWalletView/NoWalletView.tsx
```diff
@@ -0,0 +1,29 @@
...
+const NoWalletView = () => {
+  const items: Tab[] = useMemo(
+    () => [
+      { key: TabKey.TESTNET, label: 'testnet', content: <span>testnet</span> },
+      { key: TabKey.MAINNET, label: 'mainnet', content: <MainnetView /> }
+    ],
+    []
+  )
+
+  return <Tabs centerContent={true} tabs={items} defaultTabIndex={1} />
```
veado, line 26: _Just FYI (you can't know it and we have not addressed it in FIGMA yet):_ We don't need tabs to make a selection for test- or mainnet to import a wallet. Because same keystore can be used in test- and mainnet. 

See following [question / answer in our Slack history](https://thorchain.slack.com/archives/C012CLYR1DG/p1592898922420100):
Q: (4) Can an user use a same wallet (keystore) for mainnet and testnet?
A: 4) yes this is no problem. Keystores work fine for any network

src/renderer/components/wallet/NewMnemonicGenerate.tsx
```diff
@@ -4,27 +4,40 @@ import { generatePhrase } from '@thorchain/asgardex-crypto'
...
 
-  const createMnemonicWallet = () => {
-    const phrase = generatePhrase()
+  const phraseWords = useMemo(() => phrase.split(' ').map((word) => ({ text: word, _id: word })), [phrase])
+
+  const createMnemonicWallet = useCallback(() => {
     // TODO (@Veado) Extract this into helper
     localStorage.setItem('phrase', phrase)
-  }
+  }, [phrase])
```
veado, line 27: We don't need (and don't want) to store a phrase into localstorage ^

src/renderer/views/wallet/CreateView/CreateView.tsx
```diff
@@ -0,0 +1,39 @@
...
+  KEYSTORE = 'keystore'
+}
+
+export const CreateView = () => {
+  const intl = useIntl()
+  const items = useMemo(
+    () => [
+      {
+        key: TabKey.KEYSTORE,
+        label: 'keystore',
```
veado, line 21: `i18n` ?

src/renderer/views/wallet/CreateView/CreateView.tsx
```diff
@@ -0,0 +1,39 @@
...
+        content: <KeystoreView />
+      },
+      {
+        key: TabKey.PHRASE,
+        label: 'phrase',
```
veado, line 26: `i18n` ?

src/renderer/views/wallet/CreateView/PhraseView.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  if (!mnemonicInfo) {
+    return <MnemonicGenerate onSubmit={setMnemonicInfo} />
+  }
+
+  return (
+    <MnemonicConfirmScreen
+      mnemonic={mnemonicInfo.phrase}
+      onConfirm={() => {
+        keystoreService.addKeystore(mnemonicInfo.phrase, mnemonicInfo.password).then(() => {
```
veado, line 23: Do we need `then` handler here? I guess `WalletView` would handle it by changing its routes after adding a keystore.

src/renderer/views/wallet/NoWalletView/NoWalletView.tsx
```diff
@@ -0,0 +1,23 @@
...
+  content: React.ReactNode
+}
+
+const NoWalletView = () => {
+  const items: Tab[] = useMemo(() => [{ key: TabKey.MAINNET, label: 'mainnet', content: <MainnetView /> }], [])
+
+  return <Tabs centerContent={true} tabs={items} />
+}
+
+export default NoWalletView
```
veado, line 23: Since we don't need any tabs here, I would recommend to replace all of current content with all content of `MainnetView`. Layout has been updated https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020?node-id=1310%3A0 . `MainnetView` can be removed then. This change can be done by another PR ...

src/renderer/components/wallet/NewMnemonicGenerate.tsx
```diff
@@ -40,10 +53,18 @@ const NewMnemonicGenerate: React.FC<Props> = (_: Props): JSX.Element => {
...
   )
+
+  const copyPhraseToClipborad = useCallback(() => {
+    navigator.clipboard.writeText(phrase)
+  }, [phrase])
   return (
     <>
-      <h1>Generate the new mnemonic wallet</h1>
+      <Label onClick={copyPhraseToClipborad}>{intl.formatMessage({ id: 'wallet.create.copy.phrase' })}</Label>
+      <MnemonicPhrase words={phraseWords} />
```
veado, line 63: [suggestion] I would recommend to add a "refresh" button to re-generate the phrase if the user does not agree with it. I've updated the layout at FIGMA  https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020?node-id=1310%3A0

src/renderer/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,30 @@
...
+      tabs.map(({ label, key, content }) => (
+        <ATabs.TabPane tab={<Styled.TabLabel onClick={() => setActiveTabKey(key)}>{label}</Styled.TabLabel>} key={key}>
+          <Styled.ContentWrapper centerContent={centerContent}>{content}</Styled.ContentWrapper>
+        </ATabs.TabPane>
+      )),
+    [tabs, centerContent]
+  )
+
+  return (
+    <Styled.Tabs centered={true} activeKey={activeTabKey}>
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: @veado not exactly. `centered` is a new (released about 18 days ago) to center Tabs items https://ant.design/components/tabs/#components-tabs-demo-centered. `centerContent` its a custom prop to center content itself for each tab

src/renderer/components/header/HeaderLang.tsx
```diff
@@ -24,8 +24,8 @@ type Props = {
 const HeaderLang: React.FC<Props> = (props: Props): JSX.Element => {
   const { isDesktopView, changeLocale = () => {}, locale } = props
 
-  const changeLang = useCallback(
-    ({ key }: ClickParam) => {
+  const changeLang: MenuProps['onClick'] = useCallback(
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: @veado ClickParam is a type provided by `rc-menu` package used by ant and its now totally internal with latest ant version

src/renderer/views/wallet/CreateView/PhraseView.tsx
thatStrangeGuyThorchain: @veado  unfortunately yes. when form is submitted it actually triggers `keystore$` which triggers `WalletView`. but... the history's route is still `/wallet/create` and it still renders `CreateView`. there is some logic at the `WalletView` based on 'is there a keystore?' and we can not just return any `Redirect` when `isLocked === false`. so here we have to redirect a user once to some base route. and also we have to do it exactly AFTER `keystore$` is updated  (at the then handler) and info about a keystore is available to render appropriate page

src/renderer/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { useMemo, useState } from 'react'
+
+import { Tabs as ATabs } from 'antd'
+
+import * as Styled from './Tabs.styles'
+
+type Props = {
+  defaultTabIndex?: number
+  tabs: { label: string; key: string; content: React.ReactNode }[]
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: i'd like to leave it 'private'. i believe there is no need to extract this type for outside components. in case any component wants use `Tabs` component TS compiler will check props' typings anyway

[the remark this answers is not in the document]
veado: Ok.

src/renderer/views/wallet/CreateView/PhraseView.tsx
veado: In `ImportPhrase` we do same `keystoreService.addKeystore`, but handle possible errors there: https://github.com/thorchain/asgardex-electron/blob/81dfc0f751897d74d7a704f1e5d698a80ed5d234/src/renderer/components/wallet/ImportPhrase.tsx#L76-L80 

Could you try to handle errors and progress state in `MnemonicConfirmScreen` similar as we have it in `ImportPhrase`?

src/renderer/components/wallet/NewMnemonicGenerate.tsx
```diff
@@ -1,30 +1,40 @@
...
-    } catch (err) {
-      setLoadingMsg('')
-    }
-  }, [])
+  const phraseWords = useMemo(() => phrase.split(' ').map((word) => ({ text: word, _id: word })), [phrase])
+
+  const handleFormFinish = useCallback(
+    async (formData: Store) => {
+      try {
+        setLoadingMsg('Creating wallet...')
```
veado, line 30: Use `i18n` here and at other places in `NewMnemonicGenerate`

src/renderer/views/wallet/CreateView/PhraseView.tsx
thatStrangeGuyThorchain: @veado mice idea. added error handling but not sure here about loading state. we are not blocked by network requests here and adding a keystore here is just a writing ti the file on the local machine. as it really fast i believe we may ignore it and just handle ad error. dont you think so ?

veado: And we have a call of [`encryptToKeyStore` of `asgardex-crypto`](https://gitlab.com/thorchain/asgardex-common/asgardex-crypto/-/blob/master/src/crypto.ts#L70-108) in `addKeystore, which might take some time. Agree, if it's really that fast as expected, we don't need to give any "importing ..." feedback. Just let's check it out before.

thatStrangeGuyThorchain: @veado yep. it so fast as expected even in dev mode. it switches to wallet view like Flash ⚡

veado: Cool, then let skip a `progress` message as you suggested.

src/renderer/components/wallet/NewMnemonicConfirm.tsx
```diff
@@ -56,6 +59,8 @@ const MnemonicConfirmScreen: React.FC<{ mnemonic: string; onConfirm: Function }>
   const [loadingMsg] = useState<string>('')
```
veado, line 59: Do we need still a`loadingMsg` here?

src/renderer/components/wallet/NewMnemonicConfirm.tsx
```diff
@@ -199,7 +179,7 @@ const MnemonicConfirmScreen: React.FC<{ mnemonic: string; onConfirm: Function }>
         </Form.Item>
         <Form.Item>
           <Button size="large" type="primary" htmlType="submit" block>
-            {loadingMsg || 'Confirm'}
+            {loadingMsg || intl.formatMessage({ id: 'common.confirm' })}
```
veado, line 182: ^ Similar to https://github.com/thorchain/asgardex-electron/pull/285/files#r457281840

PR #293, opened 2020-07-21
Resolve "Test app service"

### LINKED ISSUE #292: Test app service

veado:
- [x] Test `network$` + `toggleNetwork`
- [x] Test `onlineStatus$`

Closes #292

### DIFF
--- a/src/renderer/tests/util.ts
+++ b/src/renderer/tests/util.ts
@@ -1,0 +1,7 @@
+import { TestScheduler } from 'rxjs/testing'
+
+export function createScheduler() {
+  return new TestScheduler((actual, expected) => {
+    expect(expected).toEqual(actual)
+  })
+}
--- a/src/renderer/tests/util.ts
+++ b/src/renderer/tests/util.ts
@@ -1,7 +1,0 @@
-import { TestScheduler } from 'rxjs/testing'
-
-export function createScheduler() {
-  return new TestScheduler((actual, expected) => {
-    expect(expected).toEqual(actual)
-  })
-}
--- a/package.json
+++ b/package.json
@@ -122,7 +122,6 @@
     "eslint-plugin-react-hooks": "^4.0.4",
     "git-revision-webpack-plugin": "^3.0.6",
     "husky": "^4.2.5",
-    "jest-marbles": "^2.5.1",
     "lint-staged": "^10.2.11",
     "nodemon": "^2.0.4",
     "prettier": "^2.0.5",
--- a/package.json
+++ b/package.json
@@ -122,6 +122,7 @@
     "eslint-plugin-react-hooks": "^4.0.4",
     "git-revision-webpack-plugin": "^3.0.6",
     "husky": "^4.2.5",
+    "jest-marbles": "^2.5.1",
     "lint-staged": "^10.2.11",
     "nodemon": "^2.0.4",
     "prettier": "^2.0.5",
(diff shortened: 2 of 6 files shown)


src/renderer/tests/util.ts
```diff
@@ -0,0 +1,7 @@
+import { TestScheduler } from 'rxjs/testing'
```
thatStrangeGuyThorchain, line 1: how do think maybe we can extract it to the `serupTests.ts` file to use it from global ?
it might me like 
```ts
const runObservable = createScheduler().run
```

veado: Do you mean to use same `TestScheduler` everywhere? I tried to use one instance of `TestScheduler` before, but had some time frame issues [with test of `onlineStatus$`](https://github.com/thorchain/asgardex-electron/blob/07f912678b0fa452a9c0f3e2830458e1eb18b27d/src/renderer/services/app/service.test.ts#L37-L62).

thatStrangeGuyThorchain: @veado  not exactly. just put this helper to the setupTests.ts file and declare createScheduler helper globally for test ENV to avoid importing it in every test

veado: Something like this fa86f0426a5efce7fc4eca1bdc5761a1a0506048 ? 

```
Test Suites: 23 passed, 23 total
Tests:       158 passed, 158 total
```

Tests are running fine ^, but I have still `tsc` error in my editor, such as `Cannot find name 'runObservable'.` Do you know how to fix these?



Screenshot from 2020-07-22 13-55-13

thatStrangeGuyThorchain: yep. like this. 
```
diff --git a/src/setupTests.ts b/src/setupTests.ts
index 682f22f..aed8333 100644
--- a/src/setupTests.ts
+++ b/src/setupTests.ts
@@ -7,6 +7,8 @@ import { Option, isNone } from 'fp-ts/lib/Option'
 import { TestScheduler } from 'rxjs/testing'
 
 declare global {
+  const runObservable: typeof TestScheduler.prototype.run
+
   // eslint-disable-next-line no-redeclare, @typescript-eslint/no-namespace
   namespace NodeJS {
     interface Global {
diff --git a/tsconfig.json b/tsconfig.json
index 055aac3..1b5217c 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -15,6 +15,10 @@
     "jsx": "react",
     "isolatedModules": true
   },
-  "include": ["src/shared", "src/renderer"],
+  "include": [
+    "src/shared",
+    "src/renderer",
+    "src/setupTests.ts"
+  ],
   "exclude": ["src/main/"]
 }
```

veado: It [works pretty nice](https://github.com/thorchain/asgardex-electron/pull/293/commits/b597de0b3bf732c47ad17621af07f0537b8a5d0e), thanks! 

Why do we have to define `runObservable` twice (once for `declare global` and once for `NodeJS.Global`?

thatStrangeGuyThorchain: @veado
first one is for global TS context to say to it that there is a `runObservable` (just like a `window` or anything from `/types/declarations.d.ts`). here we use module augmentation feature of TS and mix runObservable to global TS context.
the second one is for NodeJs' env and to be able to set to the `global` (which is a global context in terms of node's test env) `runObservable` itself

PR #294, opened 2020-07-23
Add tests for i18n service

veado:
- [x] Tests for `locale$`, `changeLocale`, `initialLocale`
- [x] Fix `runObservable`

### DIFF
--- a/src/renderer/services/i18n/service.ts
+++ b/src/renderer/services/i18n/service.ts
@@ -5,9 +5,9 @@ import { observableState } from '../../helpers/stateHelper'
 import { getLocaleFromString } from '../../i18n'
 import { Locale } from '../../i18n/types'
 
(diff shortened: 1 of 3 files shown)


src/renderer/services/i18n/service.ts
```diff
@@ -5,11 +5,11 @@ import { observableState } from '../../helpers/stateHelper'
 import { getLocaleFromString } from '../../i18n'
 import { Locale } from '../../i18n/types'
 
-const LOCALE_KEY = 'asgdx-locale'
+export const LOCALE_KEY = 'asgdx-locale'
 
-const defaultLocale = () => {
+export const defaultLocale = () => {
```
veado, line 10: _legacy code_ - no `export` needed is needed here ^

PR #296, opened 2020-07-23
Add tests for wallet service

thatStrangeGuyThorchain:


src/renderer/services/wallet/service.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      keystoreService
+        .unlock(some(none), password)
+        .catch(() => {
+          failed = true
+        })
+        .then(() => {
+          expect(failed).toBeTruthy()
+          done()
+        })
```
veado, line 80: One line here and in other tests before  (if you like)
```
    it('should reject if keystore is not imported', async () => {
      await expect(keystoreService.unlock(some(none), password)).rejects.toBeTruthy()
    })
```

src/renderer/services/wallet/service.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    })
+
+    it('should call setKeystoreState', async (done) => {
+      fsPathExistsSpy.mockImplementationOnce(() => Promise.resolve(true))
+      fsReadJSONSpy.mockImplementationOnce(() => Promise.resolve(mockDecrypt))
+
+      await keystoreService.unlock(some(some({ phrase: 'phrase' })), password)
+
+      expect(keystoreService._setKeystoreState).toBeCalledWith(some(some(mockDecrypt)))
```
veado, line 89: It looks like a similar test as you already have done in  `addKeystore` (line 47), isn't?

src/renderer/services/wallet/service.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    const password = 'password'
+    removeKeystoreSpy.mockImplementationOnce(() => Promise.resolve())
+
+    await keystoreService.addKeystore(phrase, password)
+
+    expect(removeKeystoreSpy).toBeCalled()
+    expect(fsEnsureFileSpy).toBeCalledWith(KEY_FILE)
+    expect(fsWriteJsonSpy).toBeCalledWith(KEY_FILE, mockEncrypt)
+    expect(_setKeystoreStateSpy).toBeCalledWith(some(some({ phrase })))
+    done()
```
veado, line 49: Not sure if we really need to check internal things such as "is called with `xyz`". IMO more important for tests are results of a "black box", but not how it comes to its result.

So just a check if `unlock` could be "resolved" in this case ^:
```
// Promise<void> resolves in `undefined`
 await expect(keystoreService.unlock(some(some({ phrase: 'phrase' })), password)).resolves.toBeUndefined()
``` 
And after that we could check other expected states, for example `KeystoreState`.

src/renderer/services/wallet/service.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    fsRemoveSpy.mockImplementationOnce(jest.fn)
+    fsWriteJsonSpy.mockImplementationOnce(jest.fn)
+    fsEnsureFileSpy.mockImplementationOnce(jest.fn)
+  })
+
+  it('removeKeystore', async (done) => {
+    await removeKeystore()
+    expect(fsRemoveSpy).toBeCalledWith(KEY_FILE)
+    expect(_setKeystoreStateSpy).toBeCalledWith(none)
+    done()
```
veado, line 34: What about to check `KeystoreState` here?

src/renderer/services/wallet/service.test.ts
veado: UPDATE: All `toBeCalled` for file actions (remove/fileExists/writeJson)  do make sense for sure! What about the following: Instead of using `setKeystoreStateSpy` just test the `KeystoreState` with expected values, but no check of `_setKeystoreStateSpy.toBeCalledWith`?

src/renderer/services/wallet/service.ts
```diff
@@ -76,5 +76,6 @@ export const keystoreService: KeystoreService = {
   addKeystore,
   removeKeystore,
   lock: removePhrase,
-  unlock: addPhrase
+  unlock: addPhrase,
+  _setKeystoreState: setKeystoreState
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: in general we dont need `_setKeystoreState` itself to be exported and need this only for testing purposes. usually private properties starts with `_` to show that this field is private and should not be used for any purpose from the outside

src/renderer/services/wallet/service.ts
```diff
@@ -22,11 +22,11 @@ const { get$: getKeystoreState$, set: setKeystoreState } = observableState<Keyst
 const addKeystore = async (phrase: Phrase, password: string) => {
   try {
     // remove previous keystore before adding a new one to trigger changes of `KeystoreState
-    await removeKeystore()
+    await keystoreService.removeKeystore()
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: in tests module we mock `keystoreService`'s properties and need same pointers to the functions

src/renderer/services/wallet/service.ts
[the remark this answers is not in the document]
veado: In this case I would suggest to remove `_setKeystoreState` as exported member and test against values of `KeystoreState` only to see if this state has been changed as expected.

thatStrangeGuyThorchain:
@veado 
> In some cases it make sense to check internal things (e.g. has fs.remove been called to remove a file or has fs.writeJSON been called to write something into a file. However, in general I would try to test more expected results (e.g. expected states)
i agree with you but in this case this side-effects are really important. in addition if we dont mock this nodejs-actions we will delete/create files locally which might be incorrect

src/renderer/services/wallet/service.test.ts
thatStrangeGuyThorchain: not exactly. this tests that `addPhrase` calls `setKeystoreState` with appropriate wrguments

thatStrangeGuyThorchain:
@veado i agree with your proposal to test this as a black box and test only stream result but after fighting it almost 6 hours i've lost.
https://github.com/ReactiveX/rxjs/issues/5028
LSS: its almost impossible.

veado:
> ... after fighting it almost 6 hours i've lost.
> [ReactiveX/rxjs#5028](https://github.com/ReactiveX/rxjs/issues/5028)
> LSS: its almost impossible.

Yep, that's an know issue, see ["You can't directly test RxJS code that consumes Promises..."](https://rxjs-dev.firebaseapp.com/guide/testing/marble-testing#known-issues)

We don't need to use `TestScheduler` everywhere and if needed we can just go another way: _"The solution is to test that code in isolation, with the traditional async testing methods of your testing framework."_ (see example in https://rxjs-dev.firebaseapp.com/guide/testing/marble-testing#known-issues)

We have already done it ^ in [tests for `ThemeContext`](https://github.com/thorchain/asgardex-electron/blob/81dfc0f751897d74d7a704f1e5d698a80ed5d234/src/renderer/contexts/ThemeContext.test.tsx) before: No use of `TestScheduler`, just a subscription and tests of its values. BTW: In `ThemeContext` we could use `TestScheduler` of course, but we had not introduced marble testing it at this time.

veado:
> i agree with you but in this case this side-effects are really important. in addition if we dont mock this nodejs-actions we will delete/create files locally which might be incorrect

Agree, see https://github.com/thorchain/asgardex-electron/pull/296#discussion_r459499224

src/renderer/services/wallet/service.test.ts
```diff
@@ -67,8 +58,36 @@ describe('services/keystore/service/', () => {
 
       await keystoreService.unlock(some(some({ phrase: 'phrase' })), password)
 
-      expect(keystoreService._setKeystoreState).toBeCalledWith(some(some(mockDecrypt)))
-      done()
+      keystoreService.keystore$
+        .pipe(
+          tap((val) => {
+            expect(val).toEqual(some(some(mockDecrypt)))
+            done()
```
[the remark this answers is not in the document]
veado: Nevermind, I've missed you have mocked `encryptToKeyStore`, so it does not matter what you have set before.

src/renderer/services/wallet/service.test.ts
```diff
@@ -67,8 +58,36 @@ describe('services/keystore/service/', () => {
...
+    await keystoreService.addKeystore(phrase, password)
+
+    expect(removeKeystoreSpy).toBeCalled()
+    expect(fsEnsureFileSpy).toBeCalledWith(KEY_FILE)
+    expect(fsWriteJsonSpy).toBeCalledWith(KEY_FILE, mockEncrypt)
+
+    keystoreService.keystore$
+      .pipe(
+        tap((val) => {
+          expect(val).toEqual(some(some({ phrase })))
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: im not sure here. i believe its better to have explicit values in tests. i believe the sillier test the better . also if we change `KeystoreState` in a future we will directly will see which parts of an application we should check twice

Issue #135, opened 2020-06-02
ADD: Swap screen


nomadnic:
### Design
https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020?node-id=1285%3A149

### Components
![image](https://user-images.githubusercontent.com/41812729/88872127-dd4f3000-d25c-11ea-9e98-24df9494674e.png)

### Dark Mode

![image](https://user-images.githubusercontent.com/41812729/88872161-ed670f80-d25c-11ea-9f73-2aa26e05e31f.png)

Dependent on 
- [ ] #92 
- [ ] #126 
- [ ] #129 
- [ ] #130 
- [ ] #133

thatStrangeGuyThorchain:
are we sure that such slider button is a good idea for desktop view ? its a good approach for mobiles but is it good for desktop?

thorchain-admin:
I think it's fine and unique. User needs a definite action rather than Button + Confirm modal

PR #307, opened 2020-08-03
ADD: Swap screen

thatStrangeGuyThorchain:
closes https://github.com/thorchain/asgardex-electron/issues/135


src/renderer/components/swap/CurrencyInfo/CurrrencyInfo.component.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const CurrencyInfo = ({ from, to }: CurrencyInfo) => {
+  if (!from || !to) {
+    return null
+  }
+  return (
+    <Styled.Container>
+      <div>
+        1 {from.asset.ticker} = {from.priceRune.dividedBy(to.priceRune).toFormat(5).toString()} {to.asset.ticker}
+      </div>
+      <div>slip: 1%</div>
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: there is a helper `getSwapSlip` and we may calculate it but i did not know based on which value

src/renderer/components/swap/CurrencyInfo/CurrrencyInfo.component.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export const CurrencyInfo = ({ from, to }: CurrencyInfo) => {
+  if (!from || !to) {
+    return null
+  }
+  return (
+    <Styled.Container>
+      <div>
+        1 {from.asset.ticker} = {from.priceRune.dividedBy(to.priceRune).toFormat(5).toString()} {to.asset.ticker}
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: u r absolutely right but IMO dot-notation chaining is more readable here then 
```ts
formatBn(from.priceRune.dividedBy(to.priceRune), 5).toString()
```
just because we need to 'parse' all of that braces when read the code. what do you think ?

veado:
In BEPSwap we can filter assets in dropdown

Screenshot from 2020-08-06 09-02-53

and I would suggest to do add same feature in ASGARDEX (we don't have currently).

src/renderer/components/swap/CurrencyInfo/CurrrencyInfo.component.tsx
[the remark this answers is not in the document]
veado: `formatBn(from.priceRune.dividedBy(to.priceRune), 5)` - no `toString()` is needed. 

I would prefer to use our helper functions we use everywhere. These are tested and we know what they do. If you think a `dot-notation chaining` is not readable with these functions, why not just avoid chaining here by extracting the result of the "chain" into a variable to put it into `formatBN`?

src/renderer/components/swap/Swap.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const intl = useIntl()
+  // convert to hash map here instead of using getPoolDetail
+  const poolData: Record<string, PoolData> = useMemo(() => {
+    return poolDetails.reduce((acc, cur) => {
+      if (!cur.asset) {
+        return acc
+      }
+
+      return { ...acc, [cur.asset]: toPoolData(cur) }
+    }, {} as Record<string, PoolData>)
```
veado, line 42: [suggestion] Extract such helpers (even if these small) into `utils` module or similar and test it. It looks like we can use it at other places, too.

src/renderer/components/uielements/assets/assetIcon/AssetIcon.tsx
```diff
@@ -16,9 +16,7 @@ interface Props extends React.HTMLAttributes<HTMLDivElement> {
   asset: Asset
 }
 
-const AssetIcon: React.FC<Props> = (props: Props): JSX.Element => {
-  const { asset, size = 'normal' } = props
-
+const AssetIcon: React.FC<Props> = ({ asset, size = 'normal', ...rest }): JSX.Element => {
```
veado, line 19: I would always try to avoid `...rest` - it's just unsafe.

src/renderer/components/uielements/drag/Drag.tsx
```diff
@@ -124,32 +101,26 @@ const Drag: React.FC<Props> = ({
...
-            asset={source}
-            size="small"
-          />
+          <div className="source-asset">
+            <Styled.AssetIcon asset={source} size="small" />
+          </div>
         </Draggable>
-        {title && <TitleLabel color="input">{title}</TitleLabel>}
+        {title && <Styled.TitleLabel color="input">{title}</Styled.TitleLabel>}
         <AssetIcon className="target-asset" asset={target} size="small" />
```
veado, line 122: [nitpick] Avoid to use `className` by defining a `styled-component` in `Drag.style.ts`. _I know that's not a code you have written..._

src/renderer/components/uielements/slider/Slider.tsx
```diff
@@ -15,11 +15,21 @@ const Slider: React.FC<Props> = ({
...
   return (
     <>
-      <SliderWrapper className={`slider-wrapper ${className}`} tooltipPlacement={tooltipPlacement} {...rest} />
+      <SliderWrapper
+        ref={ref}
+        className={`slider-wrapper ${className}`}
+        tooltipPlacement={tooltipPlacement}
+        getTooltipPopupContainer={getTooltipPopupContainer}
+        tipFormatter={tipFormatter}
+        {...rest}
```
veado, line 31: [suggestion] Avoid to use `...rest`. _I know we have such things in BEPSwap and just migrated these things into ASGARDEX before - but it's just unsafe..._

src/renderer/routes/swap.ts
```diff
@@ -10,10 +10,13 @@ export const base: Route<void> = {
 
 export type SwapRouteParams = { source: string; target: string }
 export const swap: Route<SwapRouteParams> = {
-  template: `${base.template}/:source-:target`,
+  /**
+   * Use '--' 'cause asset symbols have '-' separator
+   */
+  template: `${base.template}/:source--:target`,
```
veado, line 16: > Use '--' 'cause asset symbols have '-' separator

[suggestion] How about to use a `|` (pipe)? 

```
BNB.TUSDB-000--BNB.RUNE-67C
```
vs.
```
BNB.TUSDB-000|BNB.RUNE-67C
```

src/renderer/components/swap/CurrencyInfo/CurrrencyInfo.component.tsx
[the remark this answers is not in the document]
thatStrangeGuyThorchain: oh. my bad here. we do not need `toString` there. thats true

src/renderer/routes/swap.ts
thatStrangeGuyThorchain: not a problem at all :) just a convo here

src/renderer/components/swap/CurrencyInfo/CurrrencyInfo.component.tsx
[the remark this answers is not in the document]
thatStrangeGuyThorchain: so my propose is to leave/comment hardcode here and after adding the rest of business logic decide what to do here

thatStrangeGuyThorchain:
> In BEPSwap we can filter assets in dropdown
> 
> Screenshot from 2020-08-06 09-02-53
> 
> and I would suggest to do add same feature in ASGARDEX (we don't have currently).

nice idea! but lets create a separate task for filtering functionality for `AssetSelect` component?

src/renderer/components/uielements/drag/Drag.tsx
thatStrangeGuyThorchain: you are totally right but this `className` is used from the outside to set additional styles for this component. i did not want to rewrite this to use custom render prop.
but yes. u r totally right. its not clear logic here

src/renderer/views/swap/SwapView.tsx
```diff
@@ -1,28 +1,55 @@
...
+            () => <Spin size="large" />,
+            () => <span>error</span>,
+            (state) => {
+              const availableAssets = state.assetDetails
+                .filter((a) => a.asset !== undefined && !!a.asset)
+                .map((a) => ({ asset: assetFromString(a.asset as string) as Asset, priceRune: bn(a.priceRune || 0) }))
+
+              return (
+                <Swap
+                  sourceAsset={assetFromString(source.toUpperCase()) || availableAssets[0].asset}
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: same question. what should happen in this way?

src/renderer/components/uielements/assets/assetIcon/AssetIcon.tsx
thatStrangeGuyThorchain: yep but..... this is a props interface :sweat_smile:
```ts
interface Props extends React.HTMLAttributes<HTMLDivElement> {
  size?: Size
  asset: Asset
}
```
to tell the truth i would not like to pass all HTMLAtrributes :)

src/renderer/components/uielements/slider/Slider.tsx
thatStrangeGuyThorchain: same here
```ts
import { SliderSingleProps } from 'antd/lib/slider'
interface Props extends SliderSingleProps {
  className?: string
  tooltipPlacement?: TooltipPlacement
  withLabel?: boolean
}
```

src/renderer/components/uielements/assets/assetIcon/AssetIcon.tsx
veado: This is an artifact migrated from BEPSwap. We don't need to pass all `HTMLAttributes` into `AssetIcon`, maybe none of then.

src/renderer/components/uielements/drag/Drag.tsx
veado: To avoid same mistakes we have made in BEPSwap, we should try to avoid such things if we see such things...

src/renderer/components/uielements/slider/Slider.tsx
veado: Yep, that's an artifact of BEPSwap as well. Check out `SliderWrapper` and you will see there is no need to pass any other props. And if there is a need, anyone can create a custom styled component from `Slider`. 

IMO using spread operators for props and pushing "unknown" props from it into nested components is not a good idea. Imagine the case for "hidden" props (hidden in a sense of reading the code), which changes the behavior of the nested components you don't want.

veado:
> nice idea! but lets create a separate task for filtering functionality for AssetSelect component?

Yep -> https://github.com/thorchain/asgardex-electron/issues/325

PR #331, opened 2020-08-09
Issue 306/example

thatStrangeGuyThorchain:


veado:
@thatStrangeGuyThorchain Thanks for this helpful suggestion! As discussed, we might come back to a similar solution w/ another PR if needed. _Closing this PR now..._

Issue #334, opened 2020-08-10
Phrase and password are "undefined" in the PhraseConfirmScreen


bluegarnet10:
Phrase and password are "undefined" in the PhraseConfirmScreen and it's redirected to the PhraseGenerateScreen again.
https://gofile.io/d/a29NpI

<img width="1591" alt="Screen Shot 2020-08-10 at 10 20 51 PM" src="https://user-images.githubusercontent.com/30181488/89816864-53ef0600-db58-11ea-9f0a-f3f0b7106728.png">

veado:
Resolved by #335

PR #335, opened 2020-08-10
fix pass phrase in the history state

bluegarnet10:
#334


veado:
NIce catch - thx for this PR!

bluegarnet10:
It’s my pleasure!

PR #341, opened 2020-08-12
Add tests for `fpHelpers`

veado:


src/renderer/helpers/fpHelpers.test.ts
```diff
@@ -0,0 +1,64 @@
...
+describe('helpers/envHelper/', () => {
+  describe('sequenceTOptionFromArray', () => {
+    it('returns none for an empty list', () => {
+      const result = FP.pipe([], sequenceTOptionFromArray, O.isNone)
+      expect(result).toBeTruthy()
+    })
+
+    it('returns none for a list with none inside', () => {
+      const result = FP.pipe([O.none, O.some(1)], sequenceTOptionFromArray, O.isNone)
+      expect(result).toBeTruthy()
```
thatStrangeGuyThorchain, line 16: [nip] we can just replace with `.toBeNone()`

src/renderer/helpers/fpHelpers.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    it('Lifts all "some" values', () => {
+      const result = FP.pipe([O.some(1), O.some(2)], sequenceTOptionFromArray)
+      expect(result).toEqual(O.some([1, 2]))
+    })
+  })
+
+  describe('sequenceTRDFromArray', () => {
+    it('returns `failure` for an empty list', () => {
+      const errorCallback = () => new Error()
+      const result = FP.pipe([], sequenceTRDFromArray<Error, string>(errorCallback), RD.isFailure)
```
thatStrangeGuyThorchain, line 28: maybe its better to do smth like this ?
```ts
const errorCallback = () => 'empty'
const result = FP.pipe([], sequenceTRDFromArray<string, string>(errorCallback))
 expect(result).toEqual(RD.failure('empty'))
```
to check exact result ?

PR #351, opened 2020-08-13
Wallet: Clean up settings view #344

### LINKED ISSUE #344: Wallet: Clean up settings view
For upcoming `alpha` release we won't support everything we currently have at settings. Following content have to be disabled or removed:

- [x] (1) Disable "Export keystore"
- [x] (2) Hide ledger headline + its address + "Another device" button
- [x] (3) Hide content of Ethereum
- [x] (4) Hide content of Ethereum
- [x] (5) Add original address of Midgard - if not connected or offline show "Not connected" (Hint: we already have done it similar at the network dropdown in the header)
- [x] (6) Grab version from `package.json` by using `webpack` to provide custom `env` in our code (`$VERSION` or similar). We did same thing with `$COMMIT_HASH`)

settings_update

thatStrangeGuyThorchain:
closes #344

### DIFF
--- a/src/renderer/components/wallet/Settings.tsx
+++ b/src/renderer/components/wallet/Settings.tsx
@@ -55,6 +55,7 @@ const UserAccounts: UserAccountType[] = [
 
 type Props = {
   network: Network
+  apiVersion?: string
   toggleNetwork?: () => void
   address: O.Option<Address>
   lockWallet?: () => void
@@ -63,7 +64,14 @@ type Props = {
 
 const Settings: React.FC<Props> = (props: Props): JSX.Element => {
   const intl = useIntl()
-  const { address, network, toggleNetwork = () => {}, lockWallet = () => {}, removeKeystore = () => {} } = props
+  const {
+    apiVersion = '',
+    address,
+    network,
+    toggleNetwork = () => {},
+    lockWallet = () => {},
+    removeKeystore = () => {}
+  } = props
 
   const removeWallet = useCallback(() => {
     removeKeystore()
@@ -133,7 +141,7 @@ const Settings: React.FC<Props> = (props: Props): JSX.Element => {
                 </StyledClientLabel>
 
                 <StyledPlaceholder>{intl.formatMessage({ id: 'setting.version' })}</StyledPlaceholder>
-                <StyledClientLabel>v1.2.3</StyledClientLabel>
+                <StyledClientLabel>v{apiVersion}</StyledClientLabel>
                 <StyledClientButton color="warning" size="big" onClick={toggleNetwork}>
                   Change to {network === Network.MAIN ? 'testnet' : 'mainnet'}
                 </StyledClientButton>
--- a/src/renderer/components/wallet/Settings.tsx
+++ b/src/renderer/components/wallet/Settings.tsx
@@ -4,6 +4,7 @@ import { StopOutlined } from '@ant-design/icons'
 import { Address } from '@thorchain/asgardex-binance'
 import { Row, Col, Button, List } from 'antd'
 import * as O from 'fp-ts/lib/Option'
+import { pipe } from 'fp-ts/pipeable'
 import { useIntl } from 'react-intl'
 
 import { ReactComponent as UnlockOutlined } from '../../assets/svg/icon-unlock-warning.svg'
@@ -62,7 +63,7 @@ type Props = {
 
 const Settings: React.FC<Props> = (props: Props): JSX.Element => {
   const intl = useIntl()
-  const { network, toggleNetwork = () => {}, lockWallet = () => {}, removeKeystore = () => {} } = props
+  const { address, network, toggleNetwork = () => {}, lockWallet = () => {}, removeKeystore = () => {} } = props
 
   const removeWallet = useCallback(() => {
     removeKeystore()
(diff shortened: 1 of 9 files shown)


src/renderer/components/wallet/Settings.tsx
```diff
@@ -148,10 +131,17 @@ const Settings: React.FC<Props> = (props: Props): JSX.Element => {
...
               <Col span={24}>
-                <StyledPlaceholder>{intl.formatMessage({ id: 'setting.midgard' })}</StyledPlaceholder>
-                <StyledClientLabel>128.128.128.128:8080</StyledClientLabel>
+                <StyledPlaceholder>{network === Network.MAIN ? 'Binance chain' : 'Midgard API'}</StyledPlaceholder>
+
+                <StyledClientLabel>
+                  {pipe(
+                    address,
+                    O.getOrElse(() => intl.formatMessage({ id: 'setting.notconnected' }))
+                  )}
```
veado, line 140: ^ Here the url of Midgard (for `main` and `testnet`) should by displayed (not an `Address`).

src/renderer/components/wallet/Settings.tsx
```diff
@@ -148,10 +131,17 @@ const Settings: React.FC<Props> = (props: Props): JSX.Element => {
           <StyledCard>
             <Row>
               <Col span={24}>
-                <StyledPlaceholder>{intl.formatMessage({ id: 'setting.midgard' })}</StyledPlaceholder>
-                <StyledClientLabel>128.128.128.128:8080</StyledClientLabel>
+                <StyledPlaceholder>{network === Network.MAIN ? 'Binance chain' : 'Midgard API'}</StyledPlaceholder>
```
veado, line 134: The title is still `MIdgard` (`setting.midgard`), but not `Binance chain` in case of `mainnet`.

src/renderer/components/wallet/Settings.tsx
```diff
@@ -42,44 +42,20 @@ const UserAccounts: UserAccountType[] = [
...
-      {
-        name: 'Ledger',
-        address: '0x910286F93b230E221384844C4ae18a14c474E74E',
-        type: 'external'
       }
+      // {
+      //   name: 'Ledger',
+      //   address: 'tbnb1vxutrxadm0utajduxfr6wd9kqfalv0dg2wnx5y',
+      //   type: 'external'
+      // }
```
veado, line 50: Still needed ^ ?

src/renderer/components/wallet/Settings.tsx
```diff
@@ -42,44 +42,20 @@ const UserAccounts: UserAccountType[] = [
         name: 'Main',
         address: 'tbnb1vxutrxadm0utajduxfr6wd9kqfalv0dg2wnx5y',
```
veado, line 43: Pls show address of Binance Account (via `address$` of `services/binance`)). I've forgotten to mention it in the descriptioin of #344)

src/renderer/components/wallet/Settings.tsx
thatStrangeGuyThorchain: hm. and what is a title in case of testnet ?

src/renderer/components/wallet/Settings.tsx
thatStrangeGuyThorchain: actually this is an endpoint address (url or ip). Seems address is another one prop :sweat_smile:

src/renderer/components/wallet/Settings.tsx
thatStrangeGuyThorchain: nope

src/renderer/components/wallet/Settings.tsx
veado: It's still `Midgard`.

src/renderer/components/wallet/Settings.tsx
veado: `Address` in props is an address of a Binance account.

src/renderer/views/wallet/SettingsView.tsx
```diff
@@ -1,33 +1,95 @@
...
+                name: 'Main',
+                address,
+                type: 'internal'
+              }
+            ]
+          }))
+        )
+      ),
+    [_binanceAddress$]
+  )
```
veado, line 44: Maybe `useSubscription` instead of `useMemo`? `useMemo` does not make sense, since the stream `address$` should never be changed (only its values).

src/renderer/views/wallet/SettingsView.tsx
```diff
@@ -1,33 +1,95 @@
...
+
+  const onlineStatus = useObservableState<OnlineStatus>(onlineStatus$, OnlineStatus.OFF)
 
   const network = useObservableState(network$, Network.TEST)
-  const address = useObservableState(address$, O.none)
+
+  const address$ = useMemo(
+    () => (network === Network.MAIN ? binanceEndpoint$ : pipe(midgardEndpoint$, RxOperators.map(O.fromEither))),
+    [network, binanceEndpoint$, midgardEndpoint$]
+  )
```
veado, line 59: `address$` (or better `url`) it's still the url of Midgard for test AND mainnet (but not for Binance)

src/renderer/views/wallet/SettingsView.tsx
thatStrangeGuyThorchain: why `useSubscription` ?if i understand correctly it useful when you need some callbacks as it take callbacks for next and error.
`useMemo` here makes sense 'cause without it unfortunatelly there is cycled loop

veado: Ok.

PR #358, opened 2020-08-18
Add `SuccessView`

veado:
## Standard
Screenshot from 2020-08-18 08-50-02

## With "action" button (optional)

Screenshot from 2020-08-18 08-50-15


thatStrangeGuyThorchain:
why we dont use https://ant.design/components/result component here?

veado:
> why we dont use https://ant.design/components/result component here?

Good point! I'm going to update this PR to use `Result`, but with our custom icon. And we can do the same for `ErrorView` - cool!

PR #359, opened 2020-08-18
Add props: `path` to BackLink + loading to `button

veado:


src/renderer/components/uielements/backLink/BackLink.tsx
```diff
@@ -9,14 +9,19 @@ import { BackLinkWrapper, ButtonText } from './BackLink.style'
...
 }
 
-const BackLink: React.FC<Props> = ({ label, style }): JSX.Element => {
+const BackLink: React.FC<Props> = ({ label, style, path }): JSX.Element => {
   const history = useHistory()
   const intl = useIntl()
 
   const clickHandler = () => {
-    history.goBack()
+    if (path) {
```
thatStrangeGuyThorchain, line 20: if im not mistaken we cann call `history.push` directly as
```ts
history.push(path)
```

Issue #353, opened 2020-08-17
Wallet: Update Send 


veado:
- Use `validateAddress` of BinanceClient from `asgardex-binance`
- ~~Use `CoinInputAdvanced` instread of "standard" `Input` from `antd`~~ _Will be fixed with #357_ 
- Update loading state similar to freeze / unfreeze

veado:
Resolved by #366 + #370

PR #340, opened 2020-08-12
Freeze / unfreeze (RUNE only)

veado:
Closes #312

PR #366, opened 2020-08-18
Update Send

veado:
by extracting changes from #340 

As part of #353


src/renderer/services/binance/transaction.ts
```diff
@@ -1,61 +1,59 @@
...
+    ),
+    map(({ result }) => O.fromNullable(result)),
+    map((transfers) => RD.fromOption(transfers, () => Error('Transaction: empty response'))),
+    map((transfers) =>
+      FP.pipe(
+        transfers,
+        RD.map(A.head),
+        RD.chain((transfer) => RD.fromOption(transfer, () => Error('Transaction: no results received')))
+      )
+    ),
```
thatStrangeGuyThorchain, line 47: now with `LiveData` we can abstract of inner `RD` in a stream and reolace this with just 2 lines. just take a look on how much pretty it is :) 
```typescript
liveData.map(A.head),
liveData.chain(liveData.fromOption(() => Error('Transaction: no results received'))),
```

src/renderer/services/binance/transaction.ts
```diff
@@ -1,61 +1,59 @@
...
-      catchError((e) => {
-        return Rx.of(RD.failure(e))
-      }),
-      startWith(RD.pending),
-      tap(setTransaction)
-    )
-    .subscribe()
-}
+const pushTx = (clientState$: ClientState) => ({ to, amount, asset, memo }: SendTxParams) =>
+  tx$({ clientState$, to, amount, asset, memo }).subscribe(setTxRD)
```
thatStrangeGuyThorchain, line 53: nice 👍

src/renderer/services/binance/utils.ts
```diff
@@ -81,3 +86,40 @@ export const bncSymbolToAsset = (symbol: string): O.Option<Asset> =>
...
+            O.map<Asset, AssetWithBalance>((asset) => ({
+              asset,
+              balance: assetAmount(balance.free),
+              frozenBalance: assetAmount(balance.frozen)
+            }))
+          )
+        ),
+        sequenceTOptionFromArray
+      )
+    ),
```
thatStrangeGuyThorchain, line 107: seems we can avoid one pipe nesting
```ts
RD.map(
      A.map((balance) =>
        FP.pipe(
          bncSymbolToAsset(balance.symbol),
          O.map<Asset, AssetWithBalance>((asset) => ({
            asset,
            balance: assetAmount(balance.free),
            frozenBalance: assetAmount(balance.frozen)
          }))
        )
      )
    ),
    RD.map(sequenceTOptionFromArray),
```

src/renderer/services/binance/utils.ts
```diff
@@ -81,3 +86,40 @@ export const bncSymbolToAsset = (symbol: string): O.Option<Asset> =>
...
+  balancesRD: AssetsWithBalanceRD,
+  asset: O.Option<Asset>
+): O.Option<AssetWithBalance> =>
+  FP.pipe(
+    asset,
+    O.chain((asset) =>
+      FP.pipe(balancesRD, RD.map(A.findFirst((assetWB) => assetWB.asset.symbol === asset.symbol)), RD.toOption)
+    ),
+    O.flatten
+  )
```
thatStrangeGuyThorchain, line 125: what do think about such implementation ? 
```ts
export const getAssetWithBalance = (
  balancesRD: AssetsWithBalanceRD,
  asset: O.Option<Asset>
): O.Option<AssetWithBalance> =>
  FP.pipe(
    sequenceTOption(asset, RD.toOption(balancesRD)),
    O.chain(([asset, balances]) =>
      FP.pipe(
        balances,
        A.findFirst((balance) => balance.asset.symbol === asset.symbol)
      )
    )
  )
```
in this case we are always in terms of Option only

thatStrangeGuyThorchain (commented):
just discussion points

src/renderer/components/wallet/txs/SendForm.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { AssetWithBalance, AssetsWithBalanceRD, AssetsWithBalance } from '../../../services/binance/types'
+import { Input, InputNumber } from '../../uielements/input'
+import AccountSelector from './../AccountSelector'
+import * as Styled from './Send.style'
+
+type Props = {
+  assetsWB: AssetsWithBalanceRD
+  assetWB: AssetWithBalance
+  onSubmit: ({ to, amount, asset, memo }: SendTxParams) => void
+  isLoading: boolean
```
thatStrangeGuyThorchain, line 22: do we really want to use loading flags ? i mean what is a sense to handle form's state 'isLoading' instead of shohwing it only in case of initial (as it was) and sow `SuccessView` in case of success ?

veado: Yep, we do need this flag, since we don't show a "loading screen" anymore, We will still stay at the form while having a "loading" state. The form will show a spin in its submit button and all input fields are disabled.

thatStrangeGuyThorchain (approved):

PR #372, opened 2020-08-20
Quick fix: `fromPromise.test.ts` 

veado:
does need type annotations at some places - it seems `tsc` can't infer some type for any reason (at least on my machine ;) )

Screenshot from 2020-08-20 08-52-33
Screenshot from 2020-08-20 08-52-04

### DIFF
--- a/src/renderer/helpers/rx/fromPromise.test.ts
+++ b/src/renderer/helpers/rx/fromPromise.test.ts
@@ -4,7 +4,7 @@ import { fromPromise$ } from './fromPromise'
 
 describe('/helpers/rx/fromPromise', () => {
   it('should trigger next value in case Promise resolved', (done) => {
-    const targetStream$ = fromPromise$((val) => Promise.resolve(val), 0)
+    const targetStream$ = fromPromise$<number, number>((val) => Promise.resolve(val), 0)
 
     targetStream$(1)
     targetStream$(2)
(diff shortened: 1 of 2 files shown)


src/renderer/helpers/rx/fromPromise.ts
```diff
@@ -6,7 +6,7 @@ import { observableState } from '../stateHelper'
 export const fromPromise$ = <T, V>(
   factory: (next: V) => Promise<T>,
   defaultValue: T,
-  onError = (_: Error) => defaultValue
+  onError: (_: Error) => T = (_) => defaultValue
```
veado, line 9: ^ Not needed to fix type issues in `tests` - just to be more explicit in type annotations

src/renderer/helpers/rx/fromPromise.test.ts
```diff
@@ -37,10 +37,10 @@ describe('/helpers/rx/fromPromise', () => {
     })
 
     it('should trigger onError cb ', (done) => {
-      const targetStream$ = fromPromise$(
+      const targetStream$ = fromPromise$<number, number>(
         () => Promise.reject(3),
         1,
-        (e) => e * 2
+        (_) => 6
```
thatStrangeGuyThorchain, line 43: here we check that argument was passed. imo its better to use argument in test too

thatStrangeGuyThorchain (approved):

src/renderer/helpers/rx/fromPromise.test.ts
veado: `(e) => e * 2` can't work, since we have `e: Error`

thatStrangeGuyThorchain: u r right. it's a lack of implementation typings. lets leave thing this way for now

PR #381, opened 2020-08-26
Use latest asgardex-binance@3.0.0

veado:
- [x] Address breaking changes of using `asgardex-binance@3.0.0`
- [x] Use latest `@binance-chain/javascript-sdk@4.1.1`
- [x] TxTable: Show all decimal


src/renderer/services/binance/transaction.ts
```diff
@@ -33,8 +33,8 @@ const tx$ = ({
     switchMap((r) => (O.isSome(r) ? Rx.of(r.value) : Rx.EMPTY)),
     switchMap((client) =>
       memo
-        ? Rx.from(client.vaultTx(to, amount.amount().toNumber(), symbol, memo))
-        : Rx.from(client.normalTx(to, amount.amount().toNumber(), symbol))
+        ? Rx.from(client.vaultTx({ addressTo: to, amount: amount.amount().toNumber(), asset: symbol, memo }))
```
thatStrangeGuyThorchain, line 36: just noticed that we use here number value for amount. but actually we have BigNumber and with such casting we have loose percision of user's input. its not related to PR's changes just a thoughts. What do u think ?

thatStrangeGuyThorchain (approved):

src/renderer/services/binance/transaction.ts
veado: Good point :+1:  Will change it after merging https://gitlab.com/thorchain/asgardex-common/asgardex-binance/-/merge_requests/17

Issue #385, opened 2020-08-26
Swap: Update confirmation modal


veado:
### Have

![Screenshot from 2020-08-26 17-58-42](https://user-images.githubusercontent.com/61792675/91327514-e7b20a80-e7c5-11ea-922e-7a89490ad795.png)

### Want

![Screenshot from 2020-08-26 18-03-32](https://user-images.githubusercontent.com/61792675/91327925-7888e600-e7c6-11ea-8fa9-4fe1fdd07653.png)

^ https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020

PR #396, opened 2020-08-28
Swap: Update balances in wallet after a succesfull swap #386

### LINKED ISSUE #386: Swap: Update balances in wallet after a succesfull swap
Currently swap page has not be updated with latest balances after swapping 

Screen cast: https://gofile.io/d/8VEvLW

thatStrangeGuyThorchain:
closes #386 #385


src/renderer/components/swap/Swap.tsx
```diff
@@ -264,37 +274,105 @@ export const Swap = ({
...
               onCancel={resetTx}>
-              {r.hash}
+              {e.message}
             </Modal>
-          )
+          ),
+          (r) =>
+            pipe(
+              r.state,
+              RD.map(
```
veado, line 298: [question] Why do we need to wrap everything in this pipe into a RD? Maybe I miss something, but I can't see "remote" data nor any need to have 4 states (initial, loading, error, success)  here. Isn't `Either` not enough to handle success and error states here.

src/renderer/services/binance/service.ts
```diff
@@ -368,7 +370,15 @@ const freezeFee$: Observable<FeeRD> = FP.pipe(
   shareReplay(1)
 )
 
-const transaction = createTransactionService(clientState$)
+const wsTransfer$ = pipe(
+  address$,
+  switchMap(O.fold(() => Rx.EMPTY, subscribeTransfers)),
+  RxOperators.map(RD.success),
```
veado, line 376: [question] Why do we need to wrap incoming WS events into `RD`?

src/renderer/services/binance/service.ts
```diff
@@ -368,7 +370,15 @@ const freezeFee$: Observable<FeeRD> = FP.pipe(
...
+  RxOperators.tap((state) => RD.isSuccess(state) && reloadBalances()),
```
veado, line 377: I'm not sure if we really want to reload balances with any incoming transfer event. Because it can be triggered by any transaction (even for buying chocolate in a candy store), but we don't need to refresh **all** balances if the user stays at the settings page etc. We should think about it again (no need to decide it with this PR).

veado (approved):

src/renderer/services/binance/service.ts
thatStrangeGuyThorchain: @veado u r definitely right

src/renderer/components/swap/Swap.tsx
thatStrangeGuyThorchain: hmmmmm... makes sense

thatStrangeGuyThorchain: but wait... `Either` will tell us about a transaction RESULT (error or success) but just for example with which value would you start a `wsTransfer$` stream ? when we dont know about tx status yet ?

veado: As discussed in Slack, we 've decided to simplify it ^ by using `Option` (instead of `RD`)

PR #404, opened 2020-09-01
Support chaosnet in settings

### LINKED ISSUE #402: Support chaosnet

veado:
- [x] Add menu to select `testnet` | `chaosnet` | `mainnet` in `Settings`
- [x] Quick fix for styles of menu in `AssetDetail` (dark mode)
- [x] Update `binanceNetwork$` + `wsEndpoint$` to support `chaosnet`  

Closes #402

### DIFF
--- a/src/renderer/components/wallet/Settings.style.tsx
+++ b/src/renderer/components/wallet/Settings.style.tsx
@@ -1,11 +1,11 @@
-import { Divider, Row, Col, Card, List } from 'antd'
+import * as A from 'antd'
 import styled from 'styled-components'
 import { palette } from 'styled-theme'
 
-import Button from '../uielements/button'
-import Label from '../uielements/label'
+import UIButton from '../uielements/button'
+import UILabel from '../uielements/label'
 
-export const StyledTitleWrapper = styled.div`
+export const TitleWrapper = styled.div`
   margin: 0px -8px;
   display: flex;
   justify-content: center;
@@ -14,7 +14,7 @@ export const StyledTitleWrapper = styled.div`
   min-height: 70px;
 `
 
-export const StyledTitle = styled(Label)`
+export const Title = styled(UILabel)`
   color: ${palette('text', 1)};
   text-transform: uppercase;
   font-family: 'MainFontRegular';
@@ -23,12 +23,12 @@ export const StyledTitle = styled(Label)`
   line-height: 22px;
 `
 
-export const StyledDivider = styled(Divider)`
+export const Divider = styled(A.Divider)`
   margin: 0;
   border-top: 1px solid ${palette('gray', 0)};
 `
 
-export const StyledSubtitle = styled(Label)`
+export const Subtitle = styled(UILabel)`
   margin: 10px 0px;
   color: ${palette('text', 0)};
   text-transform: uppercase;
@@ -37,7 +37,7 @@ export const StyledSubtitle = styled(Label)`
   font-size: 18px;
 `
 
-export const StyledRow = styled(Row)`
+export const Row = styled(A.Row)`
   padding: 10px 30px;
   background-color: ${palette('background', 1)};
 
@@ -46,17 +46,17 @@ export const StyledRow = styled(Row)`
   }
 `
 
-export const StyledWalletCol = styled(Col)`
+export const WalletCol = styled(A.Col)`
   width: 100%;
 `
 
-export const StyledCard = styled(Card)`
+export const Card = styled(A.Card)`
   border-radius: 5px;
   background-color: ${palette('background', 1)};
   border: 1px solid ${palette('gray', 0)};
 `
 
-export const StyledOptionCard = styled(Card)`
+export const OptionCard = styled(A.Card)`
   .ant-card-body {
     padding: 12px;
     display: flex;
@@ -67,7 +67,7 @@ export const StyledOptionCard = styled(Card)`
   }
 `
 
-export const StyledOptionLabel = styled(Label)`
+export const OptionLabel = styled(UILabel)`
   padding: 0;
   display: flex;
   align-items: center;
@@ -78,7 +78,7 @@ export const StyledOptionLabel = styled(Label)`
   min-height: 38px;
 `
 
-export const StyledButton = styled(Button)`
+export const Button = styled(UIButton)`
   font-family: 'MainFontRegular';
   text-transform: uppercase;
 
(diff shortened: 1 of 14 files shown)


src/renderer/components/wallet/Settings.style.tsx
```diff
@@ -188,3 +188,39 @@ export const StyledDeviceText = styled(Label)`
...
+  background-color: ${palette('background', 1)};
+  &:hover {
+    background-color: ${palette('background', 2)};
+  }
+`
+
+export const NetworkRow = styled(A.Row).attrs({
+  alignItems: 'center'
+})`
+  width: '150px';
```
thatStrangeGuyThorchain, line 206: maybe its better to use `min-width` ?

src/renderer/components/wallet/Settings.tsx
```diff
@@ -64,111 +45,137 @@ const Settings: React.FC<Props> = (props: Props): JSX.Element => {
...
           </Col>
         )),
         O.getOrElse(() => <></>)
       ),
     [intl, userAccounts]
   )
 
+  const changeNetworkHandler: MenuProps['onClick'] = useCallback(
+    (param) => {
+      console.log('changeNetworkHandler:', param)
```
thatStrangeGuyThorchain, line 81: seems like artifact

src/renderer/components/wallet/Settings.tsx
```diff
@@ -64,111 +45,137 @@ const Settings: React.FC<Props> = (props: Props): JSX.Element => {
...
+            </Styled.AccountCard>
           </Col>
         )),
         O.getOrElse(() => <></>)
       ),
     [intl, userAccounts]
   )
 
+  const changeNetworkHandler: MenuProps['onClick'] = useCallback(
+    (param) => {
```
thatStrangeGuyThorchain, line 80: here `param` has type any. why would not we type it. ?

src/renderer/components/wallet/Settings.tsx
```diff
@@ -64,111 +45,137 @@ const Settings: React.FC<Props> = (props: Props): JSX.Element => {
...
+      const asset = param.key as Network
+      changeNetwork(asset)
+    },
+    [changeNetwork]
+  )
+
+  const networkMenu = useMemo(
+    () => (
+      <Styled.NetworkMenu onClick={changeNetworkHandler}>
+        {(['testnet', 'chaosnet', 'mainnet'] as Network[]).map((network) => (
```
thatStrangeGuyThorchain, line 91: imo its might be useful to move this array to app-level constants. what do u think ? if we create constant like
```ts
const availableNetwork = ['testnet', 'chaosnet', 'mainnet']
```
then we can create dynamic type for `Network` like
```ts
type Network = typeof availableNetwork[0]
```

src/renderer/components/wallet/Settings.tsx
```diff
@@ -64,111 +45,137 @@ const Settings: React.FC<Props> = (props: Props): JSX.Element => {
...
-                <StyledPlaceholder>{intl.formatMessage({ id: 'setting.version' })}</StyledPlaceholder>
-                <StyledClientLabel>v{apiVersion}</StyledClientLabel>
-                <StyledClientButton color="warning" size="big" onClick={toggleNetwork}>
-                  Change to {network === 'mainnet' ? 'testnet' : 'mainnet'}
-                </StyledClientButton>
+                <Styled.Placeholder>{intl.formatMessage({ id: 'setting.version' })}</Styled.Placeholder>
+                <Styled.ClientLabel>v{apiVersion}</Styled.ClientLabel>
+                <Styled.Placeholder>{intl.formatMessage({ id: 'common.network' })}</Styled.Placeholder>
+                <Dropdown overlay={networkMenu} trigger={['click']}>
+                  <Row align="middle" style={{ width: '130px' }} onClick={(e) => e.preventDefault()}>
```
thatStrangeGuyThorchain, line 168: imo its better to avoid fixed sizes in any situation. maybe its better to set `minWidth` ?

thatStrangeGuyThorchain (approved):

src/renderer/components/wallet/Settings.tsx
veado: > to move this array to app-level constants.

Good idea! 

In case of:
```
const availableNetwork = ['testnet', 'chaosnet', 'mainnet']
type Network = typeof availableNetwork[0]
```

`Network` can be any string then. That's why I prefer to keep it as before 
```
type Network = 'testnet' | 'chaosnet' | 'mainnet'`
```
to be more type-safety.

src/renderer/components/wallet/Settings.tsx
veado: I could not figure out how to type it - `antd` is still not my friend here. Maybe you have an idea ...?

src/renderer/components/wallet/Settings.style.tsx
veado: It seems we can't do this ^, it will expand the menu items in its width (see screenshot)

Screenshot from 2020-09-01 18-23-49


But we do need this:

Screenshot from 2020-09-01 18-28-44

Let me know if you have another idea ...

src/renderer/components/wallet/Settings.tsx
veado: It seems we can't do this ^, it will expand the menu items in its width (see screenshot)

Screenshot from 2020-09-01 18-23-49


But we do need this:

Screenshot from 2020-09-01 18-28-44

Let me know if you have another idea ...

src/renderer/components/wallet/Settings.style.tsx
veado: We can remove `NetworkRow` - its not used...

src/renderer/components/wallet/Settings.tsx
thatStrangeGuyThorchain: oh gosh. nope. they have [MenuInfo interface](https://github.com/react-component/menu/blob/63571e1b845e12ffc1a46c1424f64a9ce3bc8a20/src/interface.ts#L5) but it is not exported anywhere. also `Menu` component is not generic so we can not pass a type-argument there :(

src/renderer/components/wallet/Settings.tsx
thatStrangeGuyThorchain: yes. my bad. for some reason i had a better opinion about TS :(

src/renderer/components/wallet/Settings.tsx
thatStrangeGuyThorchain: ant calculates dropdown width as 100% of a parent (`Row` component in this case). we can set additional style for Row as
```ts
style={{ display: 'inline-flex', minWidth: 130 }}
```

veado: Cool, by using `display: 'inline-flex'`  we don't need a width/minWidth anymore, thx!

PR #409, opened 2020-09-03
Issue 347/create bitcoin context

baebb:
- [x] Implements and initiates `asgardex-bitcoin`
- [x] Creates `BitcoinContext` + adds its services in `services/bitcoin/service`
- [x] Extracting common things into `service/types` and `service/utils` to avoid duplication by using generic types


veado (approved):
LTGM!

Issue #316, opened 2020-08-04
Fix default value for `selectedPricePoolAsset$` 


veado:
Symbols are different for main- / testnet:

## RUNE

- Testnet `PoolAsset.RUNE67C`
- Mainnet `PoolAsset.RUNEB1A`

## TUSDB
(^ is not a default value, but needs to be updated as well)
- Testnet `PoolAsset.TUSD5D7 = TUSD-5D7`
- Mainnet `PoolAsset.TUSDB888 = TUSDB-888`

## Links
- Testnet tokens https://testnet-dex.binance.org/api/v1/tokens?limit=1000
- Mainnet tokens https://dex.binance.org/api/v1/tokens?limit=1000

PR #418, opened 2020-09-07
Issue 407/change tusd to busd

### LINKED ISSUE #407: Change `TUSDB` to `BUSD` for price selection
- Update  `TUSDB` to `BUSD` in [CURRENCY_SYMBOLS](https://github.com/thorchain/asgardex-electron/blob/9cb47b22a68bb466273ccf5cc3dd88991286c162/src/renderer/const.ts#L12-L18) and [CURRENCY_WHEIGHTS](https://github.com/thorchain/asgardex-electron/blob/9cb47b22a68bb466273ccf5cc3dd88991286c162/src/renderer/const.ts#L22-L28)
- Symbols for `PoolAsset` has to be different for `testnet` and `mainnet` (we have not considered it yet)

thatStrangeGuyThorchain:
closes #407 #316


veado:
If we have a BUSD pool, $ BUSD should be visible in price selector and be selected as default value, but is not there (see attached screen shot). Tested on testnet / chaosnet.

Screenshot from 2020-09-07 10-21-52

src/renderer/services/const.ts
```diff
@@ -2,7 +2,31 @@ import { PricePoolAssets, PoolAsset } from '../views/pools/types'
...
 export const AVAILABLE_NETWORKS: Network[] = ['testnet', 'chaosnet', 'mainnet']
+
+const commonAssets = {
+  BNB: 'BNB.BNB',
+  ETH: 'ETH.ETH',
+  BTC: 'BTC.BTC',
+  BUSDB: 'BNB.BUSDB-000'
+}
+
+const networkPools = {
```
veado, line 17:
```
type PoolAssets = CommonAssets |  'BUSD' | 'RUNE'
type Pools = Record<PoolAssets, PoolAsset>

const networkPools: Record<Network, Pools> = {...}
```

src/renderer/services/const.ts
```diff
@@ -2,7 +2,31 @@ import { PricePoolAssets, PoolAsset } from '../views/pools/types'
...
+  testnet: {
+    ...commonAssets,
+    RUNE: PoolAsset.RUNE67C
+  },
+  mainnet: {
+    ...commonAssets,
+    RUNE: PoolAsset.RUNEB1A
+  },
+  chaosnet: {
```
veado, line 26: [suggestion] I think we can remove `chaosnet` to avoid duplication. And handle it properly in `mapNetworkToPoolAssets` then...

src/renderer/services/midgard/pools.ts
```diff
@@ -0,0 +1,181 @@
...
+    reloadPoolsState,
+    poolAddresses$
+  }
+}
+
+export const networkPools$ = pipe(network$, map(mapNetworkToPoolAssets))
+
+export const getDefaultRuneAsset = () => mapNetworkToPoolAssets(getCurrentNetworkState()).RUNE
+
+export const runeAsset$ = pipe(
```
veado, line 178: `Observable<PoolAsset>` -> `Observable<RuneAsset>` (see previous comments about `RuneAsset`

src/renderer/services/midgard/utils.ts
```diff
@@ -101,7 +102,10 @@ export const getPoolDetail = (details: PoolDetails, ticker: string): O.Option<Po
  * Converts PoolDetails to the appropriate HashMap
  * Keys of the end HasMap is PoolDetails[i].asset
  */
-export const getPoolDetailsHashMap = (poolDetails: PoolDetails): Record<string, PoolData> => {
+export const getPoolDetailsHashMap = (
+  poolDetails: PoolDetails,
+  runeAsset: PoolAsset = getDefaultRuneAsset()
```
veado, line 107: `runeAsset: PoolAsset` -> `runeAsset: RuneAsset`

veado (changes requested):

src/renderer/services/const.ts
thatStrangeGuyThorchain: imo its better to have explicit setting of values. in general the fact values are duplicated now its just a coincidence... am i wrong ?

veado: Assets (or `PoolAssets`) at `chaosnet` will be the same as for  `mainnet`.

veado (approved):
Great work!

PR #427, opened 2020-09-08
Swap: Moving slider does not change value of Input field anymore #419

### LINKED ISSUE #419: Swap: Moving slider does not change value of Input field anymore
(after changing value of input "by hand" only)

Screen cast: https://gofile.io/d/G0vK0X

thatStrangeGuyThorchain:
closes #419


veado (approved):
LGTM

PR #429, opened 2020-09-09
Optional ENV for midgard url (testnet only)

veado:
by defining `REACT_APP_MIDGARD_TESTNET_URL` in `.env`

```
REACT_APP_MIDGARD_TESTNET_URL = https://midgard.bepswap.com
```

### DIFF
--- a/src/renderer/helpers/envHelper.ts
+++ b/src/renderer/helpers/envHelper.ts
@@ -1,7 +1,13 @@
+export type ENV = string
+
+/**
+ * Type guard to check wether a value is an ENV
+ **/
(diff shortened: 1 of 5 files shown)


src/renderer/helpers/envHelper.ts
```diff
@@ -1,7 +1,13 @@
+export type ENV = string
+
+/**
+ * Type guard to check wether a value is an ENV
+ **/
+export const isEnv = (env: ENV | undefined): env is ENV => env !== undefined && env.length > 0
```
thatStrangeGuyThorchain, line 6: we can just use 
```ts
export const isEnv = (env?: ENV): env is ENV => !!env

```

thatStrangeGuyThorchain (approved):

PR #431, opened 2020-09-09
Sort wallet assets #405 / Hide wallet assets with zero balances #406

### LINKED ISSUE #405: Sort wallet assets
in following order:

- BTC (if available)
- RUNE
- BNB
- BEP2 assets (maybe separate by chain groups)

Discussion: https://thorchain.slack.com/archives/C012CLYR1DG/p1598924154087000

### LINKED ISSUE #406: Hide wallet assets with zero balances
Discussion https://thorchain.slack.com/archives/C012CLYR1DG/p1598924154087000

thatStrangeGuyThorchain:
closes #405 
closes #406

### DIFF
--- a/src/renderer/services/wallet/util.ts
+++ b/src/renderer/services/wallet/util.ts
@@ -1,5 +1,4 @@
 import { assetToString } from '@thorchain/asgardex-util'
-import { getMonoid } from 'fp-ts/Array'
 import * as A from 'fp-ts/Array'
 import * as FP from 'fp-ts/function'
 import { pipe, identity } from 'fp-ts/lib/function'
@@ -8,7 +7,14 @@ import { isSome, Option } from 'fp-ts/lib/Option'
 import * as Ord from 'fp-ts/Ord'
 
 import { ZERO_BN } from '../const'
-import { KeystoreState, KeystoreContent, Phrase, AssetWithBalance, AssetsWithBalance } from './types'
+import {
+  KeystoreState,
+  KeystoreContent,
+  Phrase,
+  AssetWithBalance,
+  AssetsWithBalance,
+  assetWithBalanceMonoid
+} from './types'
 
 export const getKeystoreContent = (state: KeystoreState): Option<KeystoreContent> => pipe(state, O.chain(identity))
 
@@ -23,8 +29,6 @@ export const hasKeystoreContent = (state: KeystoreState): boolean => isSome(getK
 export const hasImportedKeystore = (state: KeystoreState): boolean => isSome(state)
 
 export const isLocked = (state: KeystoreState): boolean => !hasImportedKeystore(state) || !hasKeystoreContent(state)
-
-export const assetWithBalanceMonoid = getMonoid<AssetWithBalance>()
 
 export const filterNullableBalances = (balances: AssetsWithBalance) => {
   return FP.pipe(
--- a/src/renderer/services/wallet/util.ts
+++ b/src/renderer/services/wallet/util.ts
@@ -6,6 +6,7 @@ import * as O from 'fp-ts/lib/Option'
 import { isSome, Option } from 'fp-ts/lib/Option'
 import { Ord } from 'fp-ts/Ord'
 
+import { groupFactory } from '../../helpers/array/group'
 import { KeystoreState, KeystoreContent, Phrase, AssetWithBalance, AssetsWithBalance } from './types'
 
 export const getKeystoreContent = (state: KeystoreState): Option<KeystoreContent> => pipe(state, O.chain(identity))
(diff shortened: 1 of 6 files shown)


src/renderer/services/wallet/util.ts
```diff
@@ -17,3 +22,40 @@ export const hasKeystoreContent = (state: KeystoreState): boolean => isSome(getK
 export const hasImportedKeystore = (state: KeystoreState): boolean => isSome(state)
 
 export const isLocked = (state: KeystoreState): boolean => !hasImportedKeystore(state) || !hasKeystoreContent(state)
+
+export const assetWithBalanceMonoid = getMonoid<AssetWithBalance>()
```
veado, line 26: [suggestion] Could you move it ^ into `src/renderer/services/wallet/types.ts` where we have `type AssetWithBalance` defined?

src/renderer/services/wallet/util.ts
```diff
@@ -17,3 +22,40 @@ export const hasKeystoreContent = (state: KeystoreState): boolean => isSome(getK
...
+
+// We will compare asset strings and they automatically
+// be grouped by their chains in alphabetic order
+const byAsset = Ord.ord.contramap(Ord.ordString, (balance: AssetWithBalance) => assetToString(balance.asset))
+
+export const sortBalances = (balances: AssetsWithBalance) => {
+  return FP.pipe(
+    balances,
+    // split array for 2 parts: sortable assets and the rest
+    A.reduce([[], []] as [AssetsWithBalance, AssetsWithBalance], (acc, cur) => {
```
veado, line 50: [suggestion] How about to reduce it into `Either<AssetsWithBalance, AssetsWithBalance>` instead into an array with two elements `[[], []]`?

src/renderer/services/wallet/util.ts
```diff
@@ -17,3 +22,40 @@ export const hasKeystoreContent = (state: KeystoreState): boolean => isSome(getK
 export const hasImportedKeystore = (state: KeystoreState): boolean => isSome(state)
 
 export const isLocked = (state: KeystoreState): boolean => !hasImportedKeystore(state) || !hasKeystoreContent(state)
+
+export const assetWithBalanceMonoid = getMonoid<AssetWithBalance>()
+
+export const filterNullableBalances = (balances: AssetsWithBalance) => {
```
veado, line 28: ^ A test for `filterNullableBalances` would be nice ;)

veado (approved):
Awesome stuff - great work!

src/renderer/services/wallet/util.ts
thatStrangeGuyThorchain: @veado yeah. i thought about it but there gonna be a semantic mismatch. imo at the `...types.ts` i suppose to see only types definitions. what do you think about it ?

veado: Agree, it's maybe not the best place there... How about to put it into `fpHelper` ? If yes, could you move [ClientStateM](https://github.com/thorchain/asgardex-electron/blob/06d09a1b5d7dc4029457917ae56361447ccad08a/src/renderer/services/types.ts#L14) (from a very previous PR) there as well, pls?

src/renderer/services/wallet/util.ts
thatStrangeGuyThorchain: yeah. i thought about it too. but actually it does not make sense in this case. anyway we need to get an array as a result for this function. in case of `Either` we would need to store `Either[]` 'cause Either can store only a single value (left OR right) and after that we would need to split that array for 2 parts again and map each part as `Either` values.
in this case here is no problems with using such array. such type definition allows to define ALL array element and attempt to read outer index will be restricted by compiler
image
all nested arrays are processed in a 'safe mode'

src/renderer/services/wallet/util.ts
thatStrangeGuyThorchain: im not sure that `fpHelpers` gonna be a good place for it too. We have `AssetWithBalance` only on terms of a `Wallet`. in case of moving it to app-level `fpHelpers` we will mix file structure dependencies and start to have a 'spaghetti' code 
as we mix dependency graph by importing smth from the down-side :(

veado: Let's make it simple and put it into `wallet/types.ts` as suggested earlier. So it's near to `AssetWithBalance` and `getMonoid` it still related to types (since `Monoid` is something on type level)

src/renderer/services/wallet/util.ts
veado: You're right, we still need two arrays ...

src/renderer/services/wallet/util.ts
thatStrangeGuyThorchain: ok. np

PR #451, opened 2020-09-11
Add `Eq` instances

veado:
- `eqAsset`
- `eqBaseAmount`
- `eqApiError`
- `eqOptonsBaseAmount`
- `eqAssetWithBalance`
- `eqAssetWithBalanceRD`
- `eqAssetsWithBalance`
- `eqAssetsWithBalanceRD`

incl. tests

### DIFF
--- a/src/renderer/helpers/fp/eq.ts
+++ b/src/renderer/helpers/fp/eq.ts
@@ -1,0 +1,29 @@
+import * as RD from '@devexperts/remote-data-ts'
+import { Asset, BaseAmount } from '@thorchain/asgardex-util'
+import * as A from 'fp-ts/lib/Array'
+import * as Eq from 'fp-ts/lib/Eq'
+import * as O from 'fp-ts/lib/Option'
+
+import { AssetWithBalance, ApiError, AssetsWithBalance } from '../../services/wallet/types'
+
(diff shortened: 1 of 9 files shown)


src/renderer/helpers/fp/eq.ts
```diff
@@ -0,0 +1,39 @@
...
+import * as A from 'fp-ts/lib/Array'
+import * as Eq from 'fp-ts/lib/Eq'
+import * as O from 'fp-ts/lib/Option'
+
+import { AssetWithBalance, ApiError, AssetsWithBalance } from '../../services/wallet/types'
+
+export const eqOString = O.getEq(Eq.eqString)
+
+export const eqAsset: Eq.Eq<Asset> = {
+  equals: (x, y) => x.symbol === y.symbol
```
thatStrangeGuyThorchain, line 12: im really not sure but here we compare only symbols. is there any chance that 2 different chains have the same symbol ?

thatStrangeGuyThorchain (approved):
the only one question about comparing asset

src/renderer/helpers/fp/eq.ts
veado: Good point, I will update it.

PR #456, opened 2020-09-14
AssetsTable: Style error + cache data

### LINKED ISSUE #453: Style errors for balances
With #449 we have introduced error handling for failures of differen `asgardex-xyz` clients. It needs to be improved from UI perspective.

## Have

balances_error

## Want

Improve (1) with:

-  i18n based message
- Style error message

### LINKED ISSUE #454: Fix: Cache previous balances while reloading
It has been removed in #449, but we do need it back for UX reason ...

veado:
- [x] Add `ErrorAlert`
- [x] Use `ErrorAlert` in `AssetTable`
- [x] Update `i18n`

Screenshot from 2020-09-14 19-24-16

Closes #453 
Closes #454

### DIFF
--- a/src/renderer/components/uielements/alert/ErrorAlert.tsx
+++ b/src/renderer/components/uielements/alert/ErrorAlert.tsx
@@ -1,0 +1,29 @@
+import React from 'react'
+
+import { InfoCircleOutlined } from '@ant-design/icons'
+import { AlertProps } from 'antd/lib/alert'
+import * as A from 'fp-ts/lib/Array'
+import * as FP from 'fp-ts/lib/function'
+import * as O from 'fp-ts/lib/Option'
+
+import * as Styled from './ErrorAlert.style'
+
+type Props = Omit<AlertProps, 'description' | 'type' | 'showIcon' | 'icon'> & { descriptions?: string[] }
+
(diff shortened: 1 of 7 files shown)


src/renderer/components/uielements/alert/ErrorAlert.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import * as Styled from './ErrorAlert.style'
+
+type Props = Omit<AlertProps, 'description' | 'type' | 'showIcon' | 'icon'> & { descriptions?: string[] }
+
+const ErrorAlert: React.FC<Props> = (props: Props): JSX.Element => {
+  const { descriptions } = props
+
+  const description = FP.pipe(
+    descriptions,
```
thatStrangeGuyThorchain, line 17: it seems we can just set default value for `descriptions` as an empty array. correct me if im wrong

thatStrangeGuyThorchain (approved):
just small question there

src/renderer/components/uielements/alert/ErrorAlert.tsx
veado: `descriptions` is optional, there is no default value.

PR #484, opened 2020-09-23
Improve multi-chain handling

veado:
- [x] Extract asset / chain related stuff into `asgardex-util`
- [x] Use [latest changes](https://gitlab.com/thorchain/asgardex-common/asgardex-util/-/merge_requests/16) of `asgardex-util` to be more type-safety in case of multi-chains
- [x] Move `SendView{Chain}` into `src/renderer/views/wallet/send`
- [x] Handle different chains by using different `SendView{Chain}` (Implementation of `SendViewBTC` will be added in a next PR)

Needed for #479


src/renderer/components/wallet/txs/SendFormBTC.txt
```diff
@@ -0,0 +1,198 @@
...
+    )
+  }, [oFeeOptions])
+
+  const renderFeeError = useMemo(() => {
+    if (!isFeeError) return <></>
+
+    const amount = FP.pipe(
+      oBnbAmount,
+      // no bnb asset == zero amount
+      O.getOrElse(() => assetAmount(0))
```
thatStrangeGuyThorchain, line 89: maybe we could use  [ZERO_ASSET_AMOUNT](https://github.com/thorchain/asgardex-electron/blob/master/src/renderer/const.ts#L62) to have a single pointer always ?

src/renderer/components/wallet/txs/SendFormBTC.txt
```diff
@@ -0,0 +1,198 @@
...
+  )
+
+  const amountValidator = useCallback(
+    async (_: unknown, value: BigNumber) => {
+      // max amount for bnb
+      const maxBnbAmount = FP.pipe(
+        sequenceTOption(oFee, oBnbAmount),
+        O.fold(
+          // Set maxAmount to zero if we dont know anything about bnb and fee amounts
+          () => bn(0),
```
thatStrangeGuyThorchain, line 126: and here we could use [ZERO_BN](https://github.com/thorchain/asgardex-electron/blob/master/src/renderer/const.ts#L58) ?

thatStrangeGuyThorchain (approved):

PR #486, opened 2020-09-24
Get rid of `EMPTY_ASSET` and other improvements 

veado:
All changes are needed for type-safety multi-chain support.

- [x] Get rid of `EMPTY_ASSET` ([it will be removed](https://gitlab.com/thorchain/asgardex-common/asgardex-util/-/issues/12) from `asgardex-util`, to provide `Asset` based on supported chains only)
- [x] Removet use `dataIndex` in table of `PoolsOverview` - it's just unsafe!
- [x] Use `AssetXYZ` provided by `asgardex-util` only
- [x] `getPoolTableRowData` returns `O.Option<PoolTableRowData>` (before `PoolTableRowData`) to get rid invalid (empty) assets
- [x] Few i18n updates
- [x] Update tests

### DIFF
--- a/src/renderer/i18n/ru/common.ts
+++ b/src/renderer/i18n/ru/common.ts
@@ -41,7 +41,9 @@ const common: CommonMessages = {
   'common.max': 'Макс.',
   'common.search': 'Поиск',
   'common.retry': 'Повторить',
-  'common.add': 'Добавить'
+  'common.add': 'Добавить',
+  'common.swap': 'Swap - RU',
+  'common.liquidity': 'Liquidity - RU'
 }
 
 export default common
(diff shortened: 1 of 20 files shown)


src/renderer/i18n/ru/common.ts
```diff
@@ -41,7 +41,9 @@ const common: CommonMessages = {
   'common.max': 'Макс.',
   'common.search': 'Поиск',
   'common.retry': 'Повторить',
-  'common.add': 'Добавить'
+  'common.add': 'Добавить',
+  'common.swap': 'Swap - RU',
+  'common.liquidity': 'Liquidity - RU'
```
thatStrangeGuyThorchain, line 46: 'common.swap': 'Обмен',
'common.liquidity': 'Ликвиидиность'

thatStrangeGuyThorchain (approved):

PR #488, opened 2020-09-24
Stake: Business logic to get / display data of "share" content #441

### LINKED ISSUE #441: Stake:  Business logic to get / display data of "share" content
Depends on #434

thatStrangeGuyThorchain:
closes #441

### DIFF
--- a/src/renderer/i18n/de/stake.ts
+++ b/src/renderer/i18n/de/stake.ts
@@ -14,7 +14,8 @@ const stake: StakeMessages = {
   'stake.poolDetails.totalStakers': 'Gesamtanzahl Stakers',
   'stake.poolDetails.returnToDate': 'Ertrag bis heute',
   'stake.wallet.add': 'Add wallet - DE',
-  'stake.wallet.connect': 'Please connect your wallet - DE'
+  'stake.wallet.connect': 'Please connect your wallet - DE',
+  'stake.pool.noStakes': "You don't have any shares in this pool - DE"
 }
 
 export default stake
--- a/src/renderer/i18n/de/stake.ts
+++ b/src/renderer/i18n/de/stake.ts
@@ -13,9 +13,9 @@ const stake: StakeMessages = {
   'stake.poolDetails.totalSwaps': 'Gesamtanzahl Swaps',
   'stake.poolDetails.totalStakers': 'Gesamtanzahl Stakers',
   'stake.poolDetails.returnToDate': 'Ertrag bis heute',
-  'stake.wallet.add': 'Add wallet - DE',
-  'stake.wallet.connect': 'Please connect your wallet - DE',
-  'stake.pool.noStakes': "You don't have any shares in this pool - DE"
+  'stake.wallet.add': 'Wallet hinzufügen',
+  'stake.wallet.connect': 'Bitte verbinde Deine Wallet',
+  'stake.pool.noStakes': 'Du hast keine Anteile in diesem Pool'
 }
 
 export default stake
--- a/src/renderer/i18n/de/stake.ts
+++ b/src/renderer/i18n/de/stake.ts
@@ -12,7 +12,9 @@ const stake: StakeMessages = {
   'stake.poolDetails.allTimeVal': 'Gesamtzeit Volumen',
   'stake.poolDetails.totalSwaps': 'Gesamtanzahl Swaps',
   'stake.poolDetails.totalStakers': 'Gesamtanzahl Stakers',
-  'stake.poolDetails.returnToDate': 'Ertrag bis heute'
+  'stake.poolDetails.returnToDate': 'Ertrag bis heute',
+  'stake.wallet.add': 'Add wallet - DE',
+  'stake.wallet.connect': 'Please connect your wallet - DE'
 }
 
 export default stake
--- a/src/renderer/services/midgard/stake.ts
+++ b/src/renderer/services/midgard/stake.ts
@@ -1,0 +1,68 @@
+import * as RD from '@devexperts/remote-data-ts'
+import { Asset, assetToString } from '@thorchain/asgardex-util'
+import * as F from 'fp-ts/function'
+import * as A from 'fp-ts/lib/Array'
+import * as O from 'fp-ts/lib/Option'
+import { pipe } from 'fp-ts/pipeable'
+import * as Rx from 'rxjs'
+import { combineLatest } from 'rxjs'
+import { catchError, map, startWith, switchMap } from 'rxjs/operators'
+
+import { sequenceTOption } from '../../helpers/fpHelpers'
+import { LiveData } from '../../helpers/rx/liveData'
+import { observableState } from '../../helpers/stateHelper'
+import { DefaultApi } from '../../types/generated/midgard/apis'
+import { StakersAssetData } from '../../types/generated/midgard/models'
+
+const createStakeService = (
+  byzantine$: LiveData<Error, string>,
+  getMidgardDefaultApi: (basePath: string) => DefaultApi
+) => {
+  const api$ = byzantine$.pipe(map(RD.map(getMidgardDefaultApi)))
+
+  const { get$: address$, set: setAddress } = observableState<O.Option<string>>(O.none)
+
+  const { get$: poolAsset$, set: setPoolAsset } = observableState<O.Option<Asset>>(O.none)
+
+  const stakes$: LiveData<Error, StakersAssetData> = pipe(
+    combineLatest([api$.pipe(map(RD.toOption)), address$, poolAsset$]),
+    map(([api, address, asset]) => sequenceTOption(api, address, asset)),
+    switchMap(
+      O.fold(
+        () => Rx.EMPTY,
+        ([api, address, asset]) => api.getStakersAddressAndAssetData({ address, asset: assetToString(asset) })
+      )
+    ),
+    map(
+      F.flow(
+        A.head,
+        O.fold(() => RD.initial, RD.success)
+      )
+    ),
+    catchError((e) => {
+      /**
+       * 404 response is returned in 2 cases:
+       * 1. Pool doesn't exist at all
+       * 2. User has no any stake units for the pool
+       * In both cases return initial state as `No Data` identifier
+       */
+      if ('status' in e && e.status === 404) {
+        return Rx.of(RD.initial)
+      }
+
+      /**
+       * In all other cases return error as is
+       */
+      return Rx.of(RD.failure(Error(e)))
+    }),
+    startWith(RD.pending)
+  )
+
+  return {
+    stakes$,
+    setAddress,
+    setPoolAsset
+  }
+}
+
+export { createStakeService }
--- a/src/renderer/views/stake/Share/ShareView.helper.tsx
+++ b/src/renderer/views/stake/Share/ShareView.helper.tsx
@@ -28,6 +28,9 @@ export const renderPending = () => (
 export const getRuneShare = (stake: StakersAssetData, pool: PoolDetail) => {
   const runeDepth = bn(pool.runeDepth || 0)
   const stakeUnits = bn(stake.stakeUnits || 0)
+  /**
+   * Default is 1 as neutral element for division
+   */
   const poolUnits = bn(pool.poolUnits || 1)
 
   return runeDepth.multipliedBy(stakeUnits).div(poolUnits)
--- a/src/renderer/views/stake/Share/ShareView.helper.tsx
+++ b/src/renderer/views/stake/Share/ShareView.helper.tsx
@@ -26,8 +26,8 @@ export const renderPending = () => (
 )
 
 export const getRuneShare = (stake: StakersAssetData, pool: PoolDetail) => {
-  const runeDepth = bn(pool.runeDepth || 0)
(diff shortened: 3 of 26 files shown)


src/renderer/i18n/de/stake.ts
```diff
@@ -12,7 +12,10 @@ const stake: StakeMessages = {
   'stake.poolDetails.allTimeVal': 'Gesamtzeit Volumen',
   'stake.poolDetails.totalSwaps': 'Gesamtanzahl Swaps',
   'stake.poolDetails.totalStakers': 'Gesamtanzahl Stakers',
-  'stake.poolDetails.returnToDate': 'Ertrag bis heute'
+  'stake.poolDetails.returnToDate': 'Ertrag bis heute',
+  'stake.wallet.add': 'Add wallet - DE',
+  'stake.wallet.connect': 'Please connect your wallet - DE',
+  'stake.pool.noStakes': "You don't have any shares in this pool - DE"
```
veado, line 18:
```
  'stake.wallet.add': 'Wallet hinzufügen',
  'stake.wallet.connect': 'Bitte verbinde Deine Wallet',
  'stake.pool.noStakes': "Du hast keine Anteile in diesem Pool"

File: src/renderer/services/midgard/stake.ts [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+        O.fold(() => RD.initial, RD.success)
+      )
+    ),
+    catchError((e) => {
+      /**
+       * 404 response is returned in 2 cases:
+       * 1. Pool doesn't exist at all
+       * 2. User has no any stake units for the pool
+       * In both cases return initial state as `No Data` identifier
+       */
```
veado, line 48: Very nice to have such comments, thx!

src/renderer/views/stake/Share/ShareView.helper.tsx
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Default is 1 as neutral element for division
+   */
+  const poolUnits = bn(pool.poolUnits || 1)
+
+  return runeDepth.multipliedBy(stakeUnits).div(poolUnits)
+}
+
+export const getAssetShare = (stake: StakersAssetData, pool: PoolDetail) => {
+  const assetDepth = bn(pool.assetDepth || 0)
```
veado, line 40: [hint] `asgardex-util` provides a helper [bnOrZero](https://gitlab.com/thorchain/asgardex-common/asgardex-util/-/blob/master/src/bn.ts#L17), which does the same. Also we can think about to add a `bnOrOne` to `asgardex-util`, just because we have such thinks here now...

src/renderer/views/stake/Share/ShareView.helper.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  return assetDepth.multipliedBy(stakeUnits).div(poolUnits)
+}
+
+export const getAssetSharePrice = (assetShare: BigNumber, price: BigNumber, priceRatio: BigNumber): BaseAmount => {
+  return baseAmount(assetShare.multipliedBy(price).multipliedBy(priceRatio))
+}
+
+export const getPoolShare = ({ stakeUnits }: StakersAssetData, { poolUnits }: PoolDetail) => {
+  return poolUnits ? bnOrZero(stakeUnits).div(poolUnits).multipliedBy(100) : ZERO_BN
```
veado, line 52: Ah, you already know `bneOrZero` :)

src/renderer/views/stake/Share/ShareView.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const { asset: assetParam } = useParams<StakeRouteParams>()
+  const { service: midgardService } = useMidgardContext()
+
+  const intl = useIntl()
+
+  const stakeData = useObservableState(midgardService.stake.stakes$, RD.initial)
+  const poolDetailedInfo = useObservableState(midgardService.pools.poolDetailedState$, RD.initial)
+  const runePriceRatio = useObservableState(midgardService.pools.priceRatio$, ONE_BN)
+  const priceSymbol = useObservableState(midgardService.pools.selectedPricePoolAssetSymbol$, '')
+  const asset = assetFromString(assetParam.toUpperCase())
```
veado, line 27: How about
```
const asset = O.fromNullable(assetFromString(assetParam.toUpperCase()))
```
?

So we can add an error message if route includes invalid asset string, but not rendering anything else. We do it similar in SendView ([code 1](https://github.com/thorchain/asgardex-electron/blob/e0277323ee23fe25c347f197bcac2d986a2b100a/src/renderer/views/wallet/send/SendView.tsx#L22), [code 2](https://github.com/thorchain/asgardex-electron/blob/e0277323ee23fe25c347f197bcac2d986a2b100a/src/renderer/views/wallet/send/SendView.tsx#L66-L77). The idea behind is to provide valid `Asset` to underlaying views only (as much as we can).

src/renderer/views/stake/Share/ShareView.tsx
```diff
@@ -0,0 +1,61 @@
...
+      ([stake, pool]) => {
+        const runeShare = helpers.getRuneShare(stake, pool)
+        const assetShare = helpers.getAssetShare(stake, pool)
+        const runeStakedShare = baseAmount(runeShare)
+        const runeStakedPrice = baseAmount(runePriceRatio.multipliedBy(runeShare))
+        const assetStakedShare = baseAmount(assetShare)
+        return (
+          <PoolShare
+            source="RUNE"
+            target={asset?.chain || ''}
```
veado, line 44: _As addition to previous comment:_ 
With a valid asset, we will always have a valid chain here - that's is important for all upcoming multi-chain handling.

src/renderer/views/stake/Share/ShareView.tsx
```diff
@@ -0,0 +1,61 @@
...
+            poolShare={helpers.getPoolShare(stake, pool)}
+            assetEarnedAmount={assetToBase(assetAmount(200))}
+            assetEarnedPrice={assetToBase(assetAmount(300))}
```
veado, line 47: Why do we have hardcoded values here ^?

src/renderer/views/stake/Share/ShareView.tsx
```diff
@@ -0,0 +1,61 @@
...
+            assetStakedShare={assetStakedShare}
+            basePriceAsset={priceSymbol}
+            loading={false}
+            assetStakedPrice={helpers.getAssetSharePrice(assetShare, bnOrZero(pool.price), runePriceRatio)}
+            runeEarnedAmount={assetToBase(assetAmount(200))}
+            runeEarnedPrice={assetToBase(assetAmount(300))}
```
veado, line 53: Same here ^?

src/renderer/views/stake/StakeView.tsx
```diff
@@ -12,28 +14,44 @@ import { useParams } from 'react-router-dom'
...
+  const hasWallet = useMemo(
+    () =>
+      pipe(
+        keystore,
+        O.flatten,
+        O.map(Boolean),
+        O.getOrElse(() => false)
+      ),
+    [keystore]
+  )
```
veado, line 43: ^ Sounds like a same state we can get with [!hasImportedKeystore](https://github.com/thorchain/asgardex-electron/blob/4686fd192bb2efc692a5ef01034a347b138ecbb9/src/renderer/services/wallet/util.ts#L29), isn't?

src/renderer/views/stake/StakeView.tsx
```diff
@@ -12,28 +14,44 @@ import { useParams } from 'react-router-dom'
...
+
+  const asset = assetFromString(assetParam.toUpperCase())
```
veado, line 45:
```
const asset = O.fromNullable(assetFromString(assetParam.toUpperCase()))
```
and "jump out" to show an error view only if asset is not valid or/and not available (similar to a previous comment).

src/renderer/views/stake/StakeView.tsx
```diff
@@ -12,28 +14,44 @@ import { useParams } from 'react-router-dom'
...
 import { PoolDetailsView } from './PoolDetails/PoolDetailsView'
+import { ShareView } from './Share/ShareView'
 
 type Props = {}
 
 const StakeView: React.FC<Props> = (_) => {
   const { asset: assetParam } = useParams<StakeRouteParams>()
   const { service: midgardService } = useMidgardContext()
+  const { keystoreService } = useWalletContext()
+  const { address$ } = useBinanceContext()
```
veado, line 30: Since we will provide other chains than Binance in the near future, address should be depending on a `chain`. The current `chain` (or better: the chain the current asset runs on) can we get from `Asset`. With this information we could create a helper `addressByChain(chain: Chain): O.Option<string>` (or similar), which provides such address. What do you think?

veado (changes requested):
Nice work! Some suggestions have been added ...

src/renderer/views/stake/Share/ShareView.helper.tsx
thatStrangeGuyThorchain: yeah. just missed

src/renderer/views/stake/Share/ShareView.tsx
thatStrangeGuyThorchain: in general i agree but.... :) there are 2 moments:
1. im not sure an Option makes sense here. we need the only check for valid data right here. 
2. we have [same check at the StakeView](https://github.com/thorchain/asgardex-electron/blob/master/src/renderer/views/stake/StakeView.tsx#L37). ShareView will never render as StakeView will return `null` right now (we should add some additional view for invalid asset. thats for sure)

src/renderer/views/stake/Share/ShareView.tsx
thatStrangeGuyThorchain: this values are for `earning` section which is:
1. out of a scope of this task
2. im not sure whether we should implement it.

so right now its just a mocked values which will be or deleted or replaced with real data in a nearest future

src/renderer/views/stake/StakeView.tsx
thatStrangeGuyThorchain: not exactly. this is like [isLocked](https://github.com/thorchain/asgardex-electron/blob/4686fd192bb2efc692a5ef01034a347b138ecbb9/src/renderer/services/wallet/util.ts#L31) value :)

src/renderer/views/stake/StakeView.tsx
thatStrangeGuyThorchain: yep. i thought about it at the beginning but forgot at the end :))

src/renderer/views/stake/StakeView.tsx
thatStrangeGuyThorchain: not sure that i understand what you mean by `FIRST`

src/renderer/views/stake/StakeView.tsx
```diff
@@ -31,16 +32,7 @@ const StakeView: React.FC<Props> = (_) => {
...
-    () =>
-      pipe(
-        keystore,
-        O.flatten,
-        O.map(Boolean),
-        O.getOrElse(() => false)
-      ),
-    [keystore]
-  )
+  const hasWallet = useMemo(() => pipe(keystore, isLocked), [keystore])
```
veado, line 35: I think it should be `!isLocked`, which means- an user has a wallet imported and it's not locked (full access to all wallet data). Or what's the idea behind "hasWallet"?

veado (changes requested):

src/renderer/views/stake/Share/ShareView.tsx
veado: to 2) If we have already parsed `Asset` from the route and have same check before, why not just pass `Asset` directly to `ShareView`? Parsing an `Asset` from (same)  route is still unsafe and does not make sense then.

src/renderer/views/stake/Share/ShareView.tsx
veado: To avoid misunderstandings, add a short comment with a link to another issue, please.

src/renderer/views/stake/StakeView.tsx
veado: Feel free to add this with another PR. Could you create an issue for it before, pls.

src/renderer/views/stake/StakeView.tsx
veado: I'm still not sure if `isLocked` is correct - see https://github.com/thorchain/asgardex-electron/pull/488#discussion_r494547861

src/renderer/views/stake/Share/ShareView.tsx
veado: We still need to make sure that `Asset` is valid and available, so that we can pass a valid chain as `target` to `PoolShare`.

src/renderer/views/stake/StakeView.tsx
thatStrangeGuyThorchain: we can not do things in this way. react registers hooks in an order they are declared and uses just an array of hooks to use appropriate hook. more details are [here](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level)

veado (approved):

PR #504, opened 2020-10-01
Refactor UI to support multi-chain txs history 

veado:
Currently we support transaction history for assets of BNC only, but we do need to support other chains as well (next will be BTC).

Needed for #501


veado:
Close in favour of #505

Issue #533, opened 2020-10-16
Liquidity: Add UI to withdraw


veado:
Layout will be very similar to BEPSwap.

### Screen shot BEPSwap

![Screenshot from 2020-10-16 12-20-49](https://user-images.githubusercontent.com/61792675/96247237-271cfc00-0faa-11eb-9fa0-41627174c7d4.png)

veado:
Duplicated #438

PR #535, opened 2020-10-16
Add WithdrawStakeView + WithdrawStake component

veado:
- Refactor folder structure + filenames to be more in common

In preparation for #533


src/renderer/components/pool/PoolDetails.tsx
```diff
@@ -103,3 +103,5 @@ export const PoolDetails: React.FC<Props> = ({
     </Styled.Container>
   )
 }
+
+export default PoolDetails
```
thatStrangeGuyThorchain, line 107: may we rethink using of default exports in the application ? in general people around suppose that this is not a good idea...
https://basarat.gitbook.io/typescript/main-1/defaultisbad

https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad

https://medium.com/@rajeshnaroth/avoid-es6-default-exports-a24142978a7a

thatStrangeGuyThorchain (approved):

src/renderer/components/pool/PoolDetails.tsx
veado: Valid point :+1:

veado: _I'm going to revert these kind of changes with next PR..._

veado: ^  https://github.com/thorchain/asgardex-electron/pull/536

PR #553, opened 2020-10-23
Load / show fees for cross-chain deposits

veado:
- [x] Add const `BASE_CHAIN` + `BASE_CHAIN_ASSET` - base chain == chain RUNE is running on (currently BNC)
- [x] Load / show `pool` and/or `base` fees (base == fees for base-chain /  pool == fees for cross-chain)
- [x] `emptyFunc` for lazy devs
- [x] Fix `sequenceT...` helpers (review suggestion)

Close ##552

### DIFF
--- a/src/renderer/components/stake/add/AddStake.tsx
+++ b/src/renderer/components/stake/add/AddStake.tsx
@@ -2,15 +2,8 @@ import React, { useCallback, useMemo, useState } from 'react'
 
 import { SyncOutlined } from '@ant-design/icons'
 import * as RD from '@devexperts/remote-data-ts'
-import {
-  Asset,
-  baseAmount,
-  BaseAmount,
-  baseToAsset,
-  formatAssetAmountCurrency,
-  PoolData
-} from '@thorchain/asgardex-util'
-import { Col, Row } from 'antd'
+import { Asset, baseAmount, BaseAmount, PoolData } from '@thorchain/asgardex-util'
+import { Col } from 'antd'
 import BigNumber from 'bignumber.js'
 import * as FP from 'fp-ts/lib/function'
 import * as O from 'fp-ts/lib/Option'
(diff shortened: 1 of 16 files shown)


src/renderer/services/chain/fees.ts
```diff
@@ -34,7 +36,21 @@ const { stream$: reloadFees$, trigger: reloadFees } = triggerStream()
 
 // reload fees
 Rx.combineLatest([selectedPoolChain$, reloadFees$])
-  .pipe(RxOp.switchMap(([oChain]) => Rx.of(FP.pipe(oChain, O.map(reloadFeesByChain)))))
+  .pipe(
+    RxOp.switchMap(([oChain]) =>
+      Rx.of(
```
thatStrangeGuyThorchain, line 41: do we really need new stream here ?
seems simple `RxOp.tap` is enough ?

src/renderer/components/stake/add/AddStake.tsx
```diff
@@ -198,13 +216,16 @@ export const AddStake: React.FC<Props> = ({
...
             <Col>
-              <Styled.ReloadFeeButton onClick={reloadFee} disabled={RD.isPending(feeRD)}>
+              <Styled.ReloadFeeButton onClick={reloadFees} disabled={RD.isPending(fees)}>
                 <SyncOutlined />
               </Styled.ReloadFeeButton>
             </Col>
             <Col>
-              <Styled.FeeLabel disabled={RD.isPending(feeRD)}>
-                {intl.formatMessage({ id: 'common.fee' })}: {feeLabel}
+              <Styled.FeeLabel disabled={RD.isPending(fees)}>
```
thatStrangeGuyThorchain, line 224: maybe we could use `loading` prop here ?

src/renderer/components/stake/add/AddStake.tsx
```diff
@@ -147,36 +146,55 @@ export const AddStake: React.FC<Props> = ({
...
-    (fee: BaseAmount) =>
+    (fee: BaseAmount, asset) =>
       formatAssetAmountCurrency({
         amount: baseToAsset(fee),
-        asset: O.toUndefined(oChainAsset),
+        asset,
         trimZeros: true
       }),
-    [oChainAsset]
+    []
```
thatStrangeGuyThorchain, line 156: as it does not depend on anything we could move it to the helper file

thatStrangeGuyThorchain (approved):

src/renderer/components/stake/add/AddStake.tsx
veado: ok

src/renderer/components/stake/add/AddStake.tsx
veado: We would lost `Fee:` text before fee amounts then.

PR #555, opened 2020-10-24
Fix deposit BTC fee for using memo

### LINKED ISSUE #549: Add deposit: Change BTC fee to base on memo

veado:
- [x] Fix deposit BTC fee for using memo
- [x] Add `isBtcChain` helper
- [x] Add tests for all `chainHelpers`

Close #549

### DIFF
--- a/src/renderer/services/chain/address.ts
+++ b/src/renderer/services/chain/address.ts
@@ -1,0 +1,46 @@
+import { Chain } from '@thorchain/asgardex-util'
+import * as FP from 'fp-ts/lib/function'
+import * as O from 'fp-ts/lib/Option'
+import * as Rx from 'rxjs'
+import * as RxOp from 'rxjs/operators'
+
+import { BASE_CHAIN } from '../../const'
+import { eqChain } from '../../helpers/fp/eq'
+import * as BNB from '../binance/service'
+import * as BTC from '../bitcoin/context'
+import { selectedPoolChain$ } from '../midgard/common'
+import { AddressRx } from './types'
+
+const addressByChain$ = (chain: Chain): AddressRx => {
+  switch (chain) {
+    case 'BNB':
+      return BNB.address$
+    case 'BTC':
+      return BTC.address$
+    case 'ETH':
+      return Rx.of(O.none)
+    case 'THOR':
+      return Rx.of(O.none)
+  }
+}
+
+/**
+ * Users wallet address for cross-chain
+ */
+const crossAddress$: AddressRx = selectedPoolChain$.pipe(
+  RxOp.switchMap((oChain) =>
+    FP.pipe(
+      oChain,
+      // x-chain only
+      O.filter((chain) => !eqChain.equals(chain, BASE_CHAIN)),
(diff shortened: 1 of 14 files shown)


src/renderer/services/chain/memo.ts
```diff
@@ -0,0 +1,34 @@
...
+import { selectedPoolAsset$ } from '../midgard/common'
+import { baseAddress$ } from './address'
+import { MemoRx } from './types'
+
+/**
+ * Stake memo
+ */
+const stakeMemo$: MemoRx = Rx.combineLatest([selectedPoolAsset$, baseAddress$]).pipe(
+  RxOp.switchMap(([oPoolAsset, oBaseAddress]) =>
+    Rx.of(
```
thatStrangeGuyThorchain, line 19: do we actually need this ? 
why not just `RxOp.map` instead of `RxOp.switchMap` ?

src/renderer/services/chain/memo.ts
```diff
@@ -0,0 +1,34 @@
...
+      FP.pipe(
+        sequenceTOption(oPoolAsset, oBaseAddress),
```
thatStrangeGuyThorchain, line 21: `sequenceTOption` will fall into `None`  value if one of element is `None`. im not sure here but what about scenario when we dont really need `oBaseAddress ` ? example
* oPoolAsset is asse from `BASE_CHAIN`
* oBaseAddress === O.none

in case of BASE_CHAIN we dont need on address anyway. 
do i miss anything ?

src/renderer/services/chain/address.ts
```diff
@@ -0,0 +1,47 @@
...
+const crossAddress$: AddressRx = selectedPoolChain$.pipe(
+  RxOp.switchMap((oChain) =>
+    FP.pipe(
+      oChain,
+      // x-chain only
+      O.filter((chain) => !eqChain.equals(chain, BASE_CHAIN)),
+      O.fold(() => Rx.of(O.none), addressByChain$)
+    )
+  ),
+  RxOp.distinctUntilChanged()
```
thatStrangeGuyThorchain, line 39: as an inner value is an ADT (in this case Option) such usage of `distinctUntilChanged` will affect nothing as there are always new pointers. this should be
```ts
RxOp.distinctUntilChanged(eqOString.equals)
```

src/renderer/services/chain/address.ts
```diff
@@ -0,0 +1,47 @@
...
+)
+
+/**
+ * Users wallet address for base chain
+ */
+const baseAddress$: AddressRx = addressByChain$(BASE_CHAIN).pipe(RxOp.distinctUntilChanged())
```
thatStrangeGuyThorchain, line 45: same here

src/renderer/services/chain/memo.ts
veado: Good questions + good point :+1: ! You are right, for deposits of base-chain assets, we don't need to add an address to memo. But in case of x-chain deposit we have to. Current implementation does not reflect all scenarios, I have fixed with 8cb51e2

thatStrangeGuyThorchain (dismissed):

PR #556, opened 2020-10-26
Helper `isBaseChain` / `isBaseChainAsset`

veado:
incl. tests

### DIFF
--- a/src/renderer/const.ts
+++ b/src/renderer/const.ts
@@ -9,7 +9,6 @@ import {
   Asset,
   assetToString,
   baseAmount,
-  Chain,
   AssetBNB
 } from '@thorchain/asgardex-util'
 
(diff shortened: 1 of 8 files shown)


src/renderer/const.ts
```diff
@@ -24,8 +24,10 @@ export const BASE_CHAIN: Chain = 'BNB'
 
 /**
  * Asset of "base" chain RUNE is currently running on
+ * AssetBNB for now, but it will be changed to `NativeRUNE` in the near future
+ * TODO(@veado): Change it if we go live with `NativeRUNE`
  */
-export const BASE_CHAIN_ASSET: Asset = getChainAsset(BASE_CHAIN)
+export const BASE_CHAIN_ASSET: Asset = AssetBNB
```
veado, line 30: `getChainAsset` causes a circular dependency - I think it's ok to set `AssetBNB` hard-coded here, similar as we have done with `BASE_CHAIN` few lines above. It will save us to re-structure things just for this line...

thatStrangeGuyThorchain (dismissed):

PR #557, opened 2020-10-26
Liquidity: Add logic to handle fees for "Withdraw Liquidity" #538

### LINKED ISSUE #538: Liquidity: Add logic to handle fees for "Withdraw Liquidity"

thatStrangeGuyThorchain:
closes #538 

- [x] Created separate UI-component Fees for displaying and formatting any array of Fees (used in Witdraw and AddStake components. Also may be used at the Send component). Renders `RD`.  caches last succeed value
- [x] Added separate UI-component ReloadButton (used in Witdraw and AddStake components)
- [x] fixed missed margins for Drag component
- [x] splitted  src/renderer/services/chain/fees.ts for 2 files: for stake and unstake
- [x] added chain/unstake servies

### DIFF
--- a/src/renderer/components/uielements/fees/Fees.tsx
+++ b/src/renderer/components/uielements/fees/Fees.tsx
@@ -36,14 +36,18 @@ export const Fees: React.FC<Props> = ({ fees, hasCrossChainFee }) => {
         RD.fold(
           () => <>{prevFeesRef.current}</>,
           () => <>{prevFeesRef.current}</>,
-          (error) => <>{error.message}</>,
+          (error) => (
+            <>
+              {intl.formatMessage({ id: 'common.error' })}: {error.message}
+            </>
+          ),
           (fees) => {
             prevFeesRef.current = fees
             return <>{fees}</>
           }
         )
       ),
-    [fees]
+    [fees, intl]
   )
 
   return (
--- a/src/renderer/components/uielements/fees/Fees.tsx
+++ b/src/renderer/components/uielements/fees/Fees.tsx
@@ -15,18 +15,13 @@ type Fee = {
 
 type Props = {
   fees: RD.RemoteData<Error, Fee[]>
+  hasCrossChainFee?: boolean
 }
 
-export const Fees: React.FC<Props> = ({ fees }) => {
+export const Fees: React.FC<Props> = ({ fees, hasCrossChainFee }) => {
   const intl = useIntl()
 
   const prevFeesRef = useRef<string>('...')
-
-  const isSingleFee = FP.pipe(
-    fees,
-    RD.map((fees) => fees.length === 1),
-    RD.getOrElse(() => false)
-  )
 
   const feesFormattedValue = useMemo(
     () =>
@@ -35,7 +30,7 @@ export const Fees: React.FC<Props> = ({ fees }) => {
         RD.map(
           FP.flow(
             A.map(formatFee),
-            A.reduce('', (acc, cur) => `${acc} ${cur}`)
+            A.reduce('', (acc, cur) => `${acc}${acc ? ' + ' : ' '}${cur}`)
           )
         ),
         RD.fold(
@@ -53,7 +48,7 @@ export const Fees: React.FC<Props> = ({ fees }) => {
 
   return (
     <>
-      {intl.formatMessage({ id: isSingleFee ? 'common.fee' : 'common.fees' })}:{feesFormattedValue}
+      {intl.formatMessage({ id: hasCrossChainFee ? 'common.fees' : 'common.fee' })}:{feesFormattedValue}
     </>
   )
 }
--- a/src/renderer/components/uielements/fees/Fees.tsx
+++ b/src/renderer/components/uielements/fees/Fees.tsx
@@ -1,0 +1,59 @@
+import React, { useRef, useMemo } from 'react'
+
+import * as RD from '@devexperts/remote-data-ts'
+import { BaseAmount, Asset } from '@thorchain/asgardex-util'
+import * as FP from 'fp-ts/function'
+import * as A from 'fp-ts/lib/Array'
+import { useIntl } from 'react-intl'
+
+import { formatFee } from './Fees.helper'
+
+type Fee = {
+  amount: BaseAmount
+  asset: Asset
+}
+
+type Props = {
+  fees: RD.RemoteData<Error, Fee[]>
+}
+
+export const Fees: React.FC<Props> = ({ fees }) => {
+  const intl = useIntl()
+
+  const prevFeesRef = useRef<string>('...')
+
+  const isSingleFee = FP.pipe(
+    fees,
+    RD.map((fees) => fees.length === 1),
+    RD.getOrElse(() => false)
+  )
+
+  const feesFormattedValue = useMemo(
+    () =>
+      FP.pipe(
+        fees,
+        RD.map(
+          FP.flow(
+            A.map(formatFee),
+            A.reduce('', (acc, cur) => `${acc} ${cur}`)
+          )
+        ),
+        RD.fold(
+          () => <>{prevFeesRef.current}</>,
+          () => <>{prevFeesRef.current}</>,
+          (error) => <>{error.message}</>,
+          (fees) => {
+            prevFeesRef.current = fees
+            return <>{fees}</>
+          }
+        )
+      ),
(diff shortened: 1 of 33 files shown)


veado:
By caching previous fees you will still see fees for x-chain deposit (BTC) while switching to base-chain deposit (BNB):

Peek 2020-10-27 19-20


Also, it seem to be broken, it gets a `Fee:Error: Request failed with status code 429`

I would recommend not to cache fees, but display a loading state (as we had before)

src/renderer/components/uielements/fees/Fees.tsx
```diff
@@ -0,0 +1,58 @@
...
+            return <>{fees}</>
+          }
+        )
+      ),
+    [fees, intl]
+  )
+
+  return (
+    <>
+      {intl.formatMessage({ id: hasCrossChainFee ? 'common.fees' : 'common.fee' })}:{feesFormattedValue}
```
veado, line 55: `hasCrossChainFee` -> fees.length > 1

src/renderer/components/uielements/fees/Fees.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { formatFee } from './Fees.helper'
+
+type Fee = {
+  amount: BaseAmount
+  asset: Asset
+}
+
+type Props = {
+  fees: RD.RemoteData<Error, Fee[]>
+  hasCrossChainFee?: boolean
```
veado, line 18: `hasCrossChainFee` is not needed - see next comment

src/renderer/components/uielements/fees/Fees.tsx
```diff
@@ -0,0 +1,58 @@
...
+        fees,
+        RD.map(
+          FP.flow(
+            A.map(formatFee),
+            A.reduce('', (acc, cur) => `${acc}${acc ? ' + ' : ' '}${cur}`)
+          )
+        ),
+        RD.fold(
+          () => <>{prevFeesRef.current}</>,
+          () => <>{prevFeesRef.current}</>,
```
veado, line 38: Caching previous fees might causes issues - see https://github.com/thorchain/asgardex-electron/pull/557#issuecomment-717437656

veado (changes requested):

src/renderer/components/uielements/fees/Fees.tsx
thatStrangeGuyThorchain: not at all. fees is RD of Array not Array of RDs. also if we get folded value of inner array type there might be inconsistent in case of pending/success states

src/renderer/views/stake/withdraw/WithdrawStakeView.tsx
```diff
@@ -33,6 +35,12 @@ export const WithdrawStakeView: React.FC<Props> = (props): JSX.Element => {
     }
   } = useMidgardContext()
 
+  const { unstakeFees$, updateUnstakePercent, updateUnstakeFeesEffect$ } = useChainContext()
+
+  useSubscription(updateUnstakeFeesEffect$)
```
veado, line 40: Nice solution to the issue we had with subscription in service before :+1:

src/renderer/services/chain/fees/unstake.ts
```diff
@@ -0,0 +1,79 @@
...
+import * as BTC from '../../bitcoin/context'
+import { selectedPoolChain$ } from '../../midgard/common'
+import { getCrossChainUnstakeMemo$ } from '../memo'
+import { FeeLD, StakeFeesLD } from '../types'
+import { reloadStakeFeesByChain } from './fees.helper'
+
+const { get$: unstakePercent$, set: updateUnstakePercent } = observableState(0)
+
+// reload fees
+const updateUnstakeFeesEffect$ = Rx.combineLatest([selectedPoolChain$, unstakePercent$]).pipe(
```
veado, line 22: Please add `debounceTime(200)` and `distinctUntilChanged` to `unstakePercent$`. It will save us to have to many requests.

src/renderer/services/chain/fees/unstake.ts
```diff
@@ -0,0 +1,79 @@
...
+import { liveData, LiveData } from '../../../helpers/rx/liveData'
+import { observableState } from '../../../helpers/stateHelper'
+import * as BNB from '../../binance/service'
+import * as BTC from '../../bitcoin/context'
+import { selectedPoolChain$ } from '../../midgard/common'
+import { getCrossChainUnstakeMemo$ } from '../memo'
+import { FeeLD, StakeFeesLD } from '../types'
+import { reloadStakeFeesByChain } from './fees.helper'
+
+const { get$: unstakePercent$, set: updateUnstakePercent } = observableState(0)
```
veado, line 19: How about to add `get: unstakePercent` , so we can get latest value of it without a need to subscribe. It will reduce subscription for example in `unstakeFeeByChain$`

src/renderer/services/chain/fees/unstake.ts
```diff
@@ -0,0 +1,79 @@
...
+  )
+)
+
+const unstakeFeeByChain$ = (chain: Chain): FeeLD => {
+  switch (chain) {
+    case 'BNB':
+      return FP.pipe(
+        unstakePercent$,
+        RxOp.switchMap(() => BNB.stakeFee$)
+      )
```
veado, line 43: I would suggest not to depend on changes of unstakePercent$ for BNC fees. BNC fee won't change that often, maybe once a year or so :)
```
return BNB.stakeFee$
```

thatStrangeGuyThorchain: @veado as we need only BASE_CHAIN fee we will not fall into BTC chain which was the only one who demanded percents. i removed percents here for now and now this is just a triggerStream without updating on withdraw percent changes

src/renderer/components/uielements/fees/Fees.tsx
veado: I still think we don't need a `hasCrossChainFee` property. The component can check itself if there is one or more fees by checking the result of fees . In case of loading / error state just show `common.fee`. It will simplify things for using this component

src/renderer/services/chain/fees/fees.helper.ts
```diff
@@ -3,6 +3,17 @@ import { Chain } from '@thorchain/asgardex-util'
...
+ * @description
+ * As some of fees' streams have shareReplay(1) resuscribing to that
+ * streams will affect nothing (instead of re-requesting data). They
+ * demand manual updating by triggering their own triggeStrams.
+ *
+ * @example
+ * updateUnstakeFeesEffect$ at ./unstake.ts
+ * @example
+ * updateStakeFeesEffect$ at ./stake.ts
+ */
```
veado, line 16: Very helpful comment!

[nitpick] typo `re-subscribing` (just a missing `b`)

src/renderer/services/chain/fees/stake.ts
```diff
@@ -20,7 +20,13 @@ import { reloadStakeFeesByChain } from './fees.helper'
 // `TriggerStream` to reload stake fees
 const { stream$: reloadStakeFees$, trigger: reloadStakeFees } = triggerStream()
 
-// reload fees
+/**
+ * reload fees
+ *
+ * Has to be used ONLY on an appopriate screen
+ * @example
+ * useSubscription(updateStakeFeesEffect$) - ONLY
```
veado, line 28: Helpful comment, too!

[ntipick] typo: `appropriate view` (just a missing `r`). Last `- ONLY` can be removed I think.

src/renderer/services/chain/fees/unstake.ts
```diff
@@ -6,74 +6,64 @@ import * as Rx from 'rxjs'
...
+// `TriggerStream` to reload unstake fees
+const { stream$: reloadUnstakeFees$, trigger: reloadUnstakeFees } = triggerStream()
+
+/**
+ * reload fees
+ *
+ * Has to be used ONLY on an appopriate screen
+ * @example
+ * useSubscription(updateUnstakeFeesEffect$) - ONLY
+ */
```
veado, line 27: [ntipick] Same typo: `appropriate view` (just a missing `r`). Last - ONLY can be removed I think.

thatStrangeGuyThorchain:
closing this PR for the sake of https://github.com/thorchain/asgardex-electron/pull/568
all further changes are there

PR #558, opened 2020-10-26
[deposit] Validate balances + fees

### LINKED ISSUE #550: Add deposit: More validation needed
- wallet balance > 0
- Deposit value > fee + wallet balances
- wallet balance > fee

### LINKED ISSUE #562: Alert: Support info, warning, success, error
Refactor `ErrorAlert` to `Alert` to support `info`, `warning,`, `success` and `error`

veado:
- [x] Validate balances for `baseChainAsset`
- [x] Validate balances for `crossChainAsset`
- [x] Validate balances for `baseChainFee`
- [x] Validate balances for `crossChainFee`
- [x] Show errors in case of invalidation ^
- [x] ~Refactor `ErrorAlert` to `Alert` (Closes #562)~ **Extracted into #563**
- [x] ~Helper `isCrossChain` / `isCrossChainAsset` (incl. tests)~ **Extracted into  #564**

Closes #550


src/renderer/components/stake/add/AddStake.tsx
```diff
@@ -23,8 +32,11 @@ type Props = {
   runeAsset: Asset
   assetPrice: BigNumber
   runePrice: BigNumber
-  assetBalance: BaseAmount
-  runeBalance: BaseAmount
+  assetBalance: O.Option<BaseAmount>
+  runeBalance: O.Option<BaseAmount>
```
thatStrangeGuyThorchain, line 36: what is a reason to change this values to Options ? in term of this file there are 2 states showing that there is no balances now:
`O.none` and `O.some(baseAmount(0))`
this is a little bit confusing to find a difference between this states and to tell the truth i can't.  below there is a check inside of `hasAssetBalance` and the same time checks based on `O.isSome(oAssetBalance)` (same for RUNEs)

thatStrangeGuyThorchain (dismissed):

src/renderer/components/stake/add/AddStake.tsx
veado: This change was needed to check balances in a right way. Before we had just `baseAmount(0)`, even if the balances were not loaded. But this will trigger a "no balance error" by showing an alert, which always happen at the first rendering of the component, if balances might not be loaded. To avoid that, we do need to wrap balances into an `O.Option` here.

src/renderer/i18n/ru/stake.ts
```diff
@@ -9,6 +9,10 @@ const stake: StakeMessages = {
   'stake.totalEarnings': 'Ваш общий доход от пула',
   'stake.add.sym': 'Add {asset} - RU',
   'stake.add.asym': 'Add {assetA} + {assetB} - RU',
+  'stake.add.error.chainFeeNotCovered': 'Needed fee of {fee} is not covered by your balance: {balance} - RU',
+  'stake.add.error.nobalances': 'No balances - RU',
+  'stake.add.error.nobalance1': "You don't have any balance of {asset} in your wallet to deposit. - RU",
+  'stake.add.error.nobalance2': "You don't have any balances of {asset1} and {asset2} in your wallet to deposit. - RU",
```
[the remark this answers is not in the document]
veado: Thanks!

PR #564, opened 2020-10-27
Add `isCrossChain` / `isCrossChainAsset` helpers

veado:
- [x] Add `isCrossChain` / `isCrossChainAsset` helpers (incl. tests)
- [x] Fix `isCrossChainStake$`

### DIFF
--- a/src/renderer/helpers/chainHelper.test.ts
+++ b/src/renderer/helpers/chainHelper.test.ts
@@ -1,6 +1,13 @@
 import { AssetBNB, AssetBTC, AssetETH, AssetRune67C, AssetRuneB1A, AssetRuneNative } from '@thorchain/asgardex-util'
 
-import { getChainAsset, isBaseChain, isBaseChainAsset, isBtcChain } from './chainHelper'
+import {
+  getChainAsset,
+  isBaseChain,
+  isBaseChainAsset,
+  isBtcChain,
+  isCrossChain,
+  isCrossChainAsset
+} from './chainHelper'
 
 describe('helpers/chainHelper', () => {
   describe('getChainAsset', () => {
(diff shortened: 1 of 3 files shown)


src/renderer/helpers/chainHelper.ts
```diff
@@ -26,6 +26,20 @@ export const isBaseChain = (chain: Chain): chain is typeof BASE_CHAIN => eqChain
...
+ */
+export const isCrossChain = (chain: Chain): boolean => !isBaseChain(chain)
+
+/**
+ * Check whether asset is a CROSS chain asset
+ */
+export const isCrossChainAsset = (asset: Asset): boolean => {
+  if (isBaseChainAsset(asset)) return false
+  const crossChainAsset = getChainAsset(asset.chain)
+  return eqAsset.equals(crossChainAsset, asset)
```
thatStrangeGuyThorchain, line 40: im not sure here but should asset ALWAYS equals to the `getChainAsset(asset.chain)` ?

thatStrangeGuyThorchain (approved):

src/renderer/helpers/chainHelper.test.ts
```diff
@@ -42,6 +49,33 @@ describe('helpers/chainHelper', () => {
     })
   })
 
+  describe('isCrossChain', () => {
+    it('false for BNB chain', () => {
+      expect(isCrossChain('BNB')).toBeFalsy()
```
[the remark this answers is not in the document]
veado: That's exactly what we want - a failing test for `BNB` if we change that constant

src/renderer/helpers/chainHelper.ts
veado: Yes, we always have one cross chain asset for each chain. If an asset is equal to this "chain asset", then it's exactly this cross chain asset.

PR #583, opened 2020-11-04
Switch to `@xchainjs/xchain-binance` #573

### LINKED ISSUE #573: Switch to  `@xchainjs/xchain-binance`
to get rid of `asgardex-binance`

Source of `@xchainjs/xchain-binance`: https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-binance

thatStrangeGuyThorchain:
closes #573


src/renderer/services/binance/service.ts
```diff
@@ -280,9 +287,18 @@ const loadAssetTxs$ = ({
   const endTime = Date.now()
   // 90 day window - similar to ASGARDEX wallet approach,
   // see https://gitlab.com/thorchain/asgard-wallet/-/blob/develop/imports/api/wallet.js#L39-48
-  const diffTime = 90 * 24 * 60 * 60 * 1000
+  const diffTime = 90 * 24 * 60 * 60 * 1001
```
veado, line 290: Why `1001`?

src/renderer/services/binance/service.ts
```diff
@@ -280,9 +287,18 @@ const loadAssetTxs$ = ({
...
   const startTime = endTime - diffTime
-  return Rx.from(client.getTransactions({ txAsset, endTime, startTime, limit, offset })).pipe(
+  return Rx.from(
+    // @TODO add txAsset to the xchain
```
veado, line 293: ^ Pls refer to a GH issue (or create one if we don't have it yet).

src/renderer/services/binance/service.ts
```diff
@@ -335,63 +351,63 @@ const assetTxs$: AssetTxsPageLD = Rx.combineLatest([
...
+  switchMap(([_, oClient]) =>
     FP.pipe(
       // client and asset has to be available
       oClient,
       // ignore all assets from other chains than BNB
       O.fold(
-        () => Rx.of(RD.initial),
-        (client) => loadFees$(client)
+        () => Rx.EMPTY,
+        (client) => Rx.from(Promise.all([client.getFees(), client.getMultiSendFees()]))
```
veado, line 379: This will end up in two requests to the same endpoint behind the scenes. Currently that's an issue of `xchain-binance` (not of this PR), but we should fix it or at least create an issue about it at `xchainjs`. 

- Request one: https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-binance/src/client.ts#L391
- Request two: https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-binance/src/client.ts#L415

veado (dismissed):
LGTM! Just few things ...

src/renderer/services/binance/service.ts
thatStrangeGuyThorchain: @veado it was a tmp TODO for my self. this is already done at the xchain package just missed this TODO comment. will just remove it in few momemnt

src/renderer/services/binance/service.ts
thatStrangeGuyThorchain: dont know :sweat_smile:
seems some local tests which i missed to revert. few minutes pls

src/renderer/services/binance/service.ts
thatStrangeGuyThorchain: agree. not sure what we should to do exactly. Do you suppose to make the only call for `getFees` and use it to provide data to the `single/multi` fees ?

veado: I would suppose to add another method to `xchain-binance`:

```ts
getSingleAndMultiFees(): Promise<{single: Fees, multi: Fees}>
```

However, I don't know if we still need `multi` Fees in ASGDX...?

veado (approved):

src/renderer/services/binance/service.ts
veado: ^ https://github.com/xchainjs/xchainjs-lib/issues/95

PR #639, opened 2020-11-30
Added network property to retrieving btc address from ledger device

sarawut11:


src/main/api/ledger.ts
```diff
@@ -1,14 +1,20 @@
...
 
-export const getBTCAddress = async () => {
+export const getBTCAddress = async (network: Network) => {
   try {
     const transport = await TransportNodeHid.open('')
     const appBtc = new AppBtc(transport)
-    const info = await appBtc.getWalletPublicKey("44'/0'/0'/0/0")
+    let info
+    if (network === 'mainnet') {
+      info = await appBtc.getWalletPublicKey("44'/0'/0'/0/0")
```
thatStrangeGuyThorchain, line 14: just a nitpick
could you pls move ` "44'/0'/0'/0/0"` and etc to the semantically named constants and description about WHY they have such values ?

src/renderer/services/bitcoin/types.ts
```diff
@@ -39,7 +39,8 @@ export type FeesService = {
 }
 
 export type LedgerService = {
-  ledgerAddress$: LedgerAddressLD
-  retrieveLedgerAddress: () => void
-  removeLedgerAddress: () => void
+  ledgerMainnetAddress$: LedgerAddressLD
+  ledgerTestnetAddress$: LedgerAddressLD
```
thatStrangeGuyThorchain, line 43: do we need them to be separated HERE ? 
lets do the same as to `retrieveLedgerAddress` and make it depends on the `network`

src/renderer/views/wallet/SettingsView.tsx
```diff
@@ -82,7 +84,13 @@ export const SettingsView: React.FC = (): JSX.Element => {
     [ethContext.address$]
   )
 
-  const bitcoinLedgerAddress = useObservableState(bitcoinContext.ledgerAddress$, RD.initial)
+  const bitcoinLedgerMainnetAddress = useObservableState(bitcoinContext.ledgerMainnetAddress$, RD.initial)
+  const bitcoinLedgerTestnetAddress = useObservableState(bitcoinContext.ledgerTestnetAddress$, RD.initial)
```
thatStrangeGuyThorchain, line 88: seems if we do this `bitcoinContext.ledgerTestnetAddress$: (network: Network) => LiveData<Error, string>`
we could use it just like 
```ts
const [bitcoinLedgerAddress] = useObservableState(() => bitcoinContext.ledgerTestnetAddress$(network), RD.initial)
```

thatStrangeGuyThorchain (changes requested):
Nice.
Just take a look at some comments

src/main/api/ledger.ts
```diff
@@ -1,14 +1,20 @@
...
+    } else {
+      info = await appBtc.getWalletPublicKey("44'/1'/0'/0/0")
+    }
```
veado, line 17: Since `chaosnet` === `mainnet` (you could not know it), the `else` would return a path for `testnet` and `chaosnet`, which is wrong.

src/renderer/components/wallet/settings/Settings.stories.tsx
```diff
@@ -17,8 +16,12 @@ storiesOf('Wallet/Settings', module).add('default', () => {
...
       lockWallet={() => console.log('lock')}
       removeKeystore={() => console.log('removeKeystore')}
-      retrieveLedgerAddress={({ chain }: LedgerAddressParams) => console.log('retrieve ledger address: ', chain)}
-      removeLedgerAddress={(chain: Chain) => console.log('remove ledger address: ', chain)}
+      retrieveLedgerAddress={({ chain, network }: LedgerAddressParams) =>
+        console.log('retrieve ledger address: ', chain, network)
+      }
+      removeLedgerAddress={({ chain, network }: LedgerAddressParams) =>
+        console.log('remove ledger address: ', chain, network)
+      }
```
veado, line 24: Do we need to pass network around to remove ledger? Changing from one network to another, the previous connected account should be removed.

src/renderer/components/wallet/settings/Settings.tsx
```diff
@@ -24,8 +24,8 @@ type Props = {
   userAccounts?: O.Option<UserAccountType[]>
   lockWallet?: () => void
   removeKeystore?: () => void
-  retrieveLedgerAddress: ({ chain }: LedgerAddressParams) => void
-  removeLedgerAddress: (chain: Chain) => void
+  retrieveLedgerAddress: ({ chain, network }: LedgerAddressParams) => void
+  removeLedgerAddress: ({ chain, network }: LedgerAddressParams) => void
```
veado, line 28: See previous comment.

src/renderer/components/wallet/settings/Settings.tsx
```diff
@@ -48,16 +48,30 @@ export const Settings: React.FC<Props> = (props): JSX.Element => {
...
   const addDevice = useCallback(
     (chain: Chain) => {
-      retrieveLedgerAddress({ chain })
+      switch (selectedNetwork) {
+        case 'mainnet':
+        case 'testnet':
+          retrieveLedgerAddress({ chain, network: selectedNetwork })
+          break
+        default:
+          break
```
veado, line 57: Why do add a switch/case statement here?

src/renderer/components/wallet/settings/Settings.tsx
```diff
@@ -48,16 +48,30 @@ export const Settings: React.FC<Props> = (props): JSX.Element => {
...
     (chain: Chain) => {
-      removeLedgerAddress(chain)
+      switch (selectedNetwork) {
+        case 'mainnet':
+        case 'testnet':
+          removeLedgerAddress({ chain, network: selectedNetwork })
+          break
+        default:
+          break
+      }
```
veado, line 72: No check of network needed (see previous comments about network + `removeLedgerAddress`)

src/renderer/contexts/BitcoinContext.tsx
```diff
@@ -23,7 +24,8 @@ export type BitcoinContextValue = {
   pushTx: typeof pushTx
   reloadFees: typeof reloadFees
   resetTx: typeof resetTx
-  ledgerAddress$: typeof ledgerAddress$
+  ledgerMainnetAddress$: typeof ledgerMainnetAddress$
+  ledgerTestnetAddress$: typeof ledgerTestnetAddress$
```
veado, line 28: You might need just one ledgerAddress$ , which depends on `network$`

src/renderer/services/bitcoin/ledger.ts
```diff
@@ -7,20 +8,42 @@ import { observableState } from '../../helpers/stateHelper'
 import { LedgerAddressRD } from '../wallet/types'
 import { LedgerService } from './types'
 
-const { get$: ledgerAddress$, set: setLedgerAddressRD } = observableState<LedgerAddressRD>(RD.initial)
+const { get$: ledgerMainnetAddress$, set: setLedgerMainnetAddressRD } = observableState<LedgerAddressRD>(RD.initial)
+const { get$: ledgerTestnetAddress$, set: setLedgerTestnetAddressRD } = observableState<LedgerAddressRD>(RD.initial)
```
veado, line 12: One `ledgerAddress$` only (see previous comment)

src/renderer/services/bitcoin/ledger.ts
```diff
@@ -7,20 +8,42 @@ import { observableState } from '../../helpers/stateHelper'
...
+  ).subscribe((value) => {
+    switch (network) {
+      case 'testnet':
+        setLedgerTestnetAddressRD(value)
+        break
+      default:
+        setLedgerMainnetAddressRD(value)
+        break
+    }
+  })
```
veado, line 29: By using `ledgerAddress$` this `switch/case` statement will be gone ^.

veado (changes requested):

src/renderer/contexts/BitcoinContext.tsx
thatStrangeGuyThorchain: @veado not exactly. i thought about it but then realized that we need to reconnect ledger advice for EVERY network changing. what do you think ?

src/renderer/components/wallet/settings/Settings.stories.tsx
thatStrangeGuyThorchain: @veado hm. why changing network should disconnect connected address ?

src/renderer/components/wallet/settings/Settings.tsx
```diff
@@ -25,7 +25,8 @@ type Props = {
   lockWallet?: () => void
   removeKeystore?: () => void
   retrieveLedgerAddress: ({ chain, network }: LedgerAddressParams) => void
-  removeLedgerAddress: ({ chain, network }: LedgerAddressParams) => void
+  removeLedgerAddress: (chain: Chain) => void
+  resetAllLedgerAddress: () => void
```
veado: Just curious: What's the reason to `remove` and / or to `reset` a Ledger?

src/renderer/components/wallet/settings/Settings.stories.tsx
veado: IMO it does not make sense to store an address for testnet in memory if you running mainnet for example.

thatStrangeGuyThorchain: @veado but it makes sense when you switch between networks. main => test => main => etc. 
about storing few additional objects in a memory when we are talking about JS.... :) im not sure that few objects in a memory is really a thing which worth to worry about :)

sarawut11: Can we keep it simple for now in this PR and continue discussion again later if this topic is raised as important one?

veado: We never store any data of another network. It's not about memory, it just does not make sense. There is a very rare use case somebody would switch between networks. You will use ASGDX to do swap, deposit, send, receive, but not to toggle between networks at first place.

thatStrangeGuyThorchain: @veado oooookay. if we dont need to care about re-connecting to the device after switching between networks that single stream is enough here.

src/renderer/contexts/BitcoinContext.tsx
veado: see https://github.com/thorchain/asgardex-electron/pull/639#discussion_r533324044

src/renderer/components/wallet/settings/Settings.tsx
veado: @sarawut11 [As discussed at Slack](https://thorchain.slack.com/archives/C012CLYR1DG/p1606813903105200?thread_ts=1606748252.098300&cid=C012CLYR1DG): Rename `reset*` to `remove*` (to avoid missunderstandings).

veado (approved):

PR #658, opened 2020-12-04
[wallet] Add BTC from Ledger to AssetOverview #645

### LINKED ISSUE #645: [wallet] Add BTC from Ledger to AssetOverview

thatStrangeGuyThorchain:
closes #645 

Main changes
 - Removed `address$` from WalletContext and store wallet address at the url.
It also fixes issue for reload page at the `SendView` and `ReceiveView` (master’s version does not restore data correctly)

- Replaced using of `xchain-client`’s `Balance` type to custom `WalletBalance` type as we need to handle balances for different addresses


veado (changes requested):
- Connecting ledger does not work for me, I got still a `b...` but not a `tb...` (testnet) address.

- Trying to add an unlocked Ledger shows an `Unknown Error` to the user (see attached screen shot) Can we change it to return a more expressive message, such as `Please unlock your Ledger` or something? 

- Naming might be confusing at some places: Since `wallet` is used as a single address (wallet can have multiple addresses), I would prefer to use `address` everywhere. In code you have already used `walletAddress` is used. What do you think?
Screenshot from 2020-12-07 08-44-52

src/renderer/services/clients/types.ts
```diff
@@ -37,15 +38,15 @@ export type LoadTxsParams = {
...
 export type TxsPageRD = RD.RemoteData<ApiError, TxsPage>
 export type TxsPageLD = LiveData<ApiError, TxsPage>
 
-export type BalanceRD = RD.RemoteData<ApiError, Balance>
+export type BalanceRD = RD.RemoteData<ApiError, WalletBalance>
 
-export type BalancesRD = RD.RemoteData<ApiError, Balances>
-export type BalancesLD = LiveData<ApiError, Balances>
+export type BalancesRD = RD.RemoteData<ApiError, WalletBalance[]>
+export type BalancesLD = LiveData<ApiError, WalletBalance[]>
```
veado, line 49: How about to rename ^`

`BalancesRD` ->  `WalletBalancesRD`
`BalancesLD` ->  `WalletBalancesLD`
?

src/renderer/services/wallet/types.ts
```diff
@@ -43,9 +44,9 @@ export type ChainBalance = {
 
 export type ChainBalances = ChainBalance[]
 
-export const BalanceMonoid = getMonoid<Balance>()
+export const BalanceMonoid = getMonoid<WalletBalance>()
 
-export type NonEmptyBalances = NonEmptyArray<Balance>
+export type NonEmptyBalances = NonEmptyArray<WalletBalance>
```
veado, line 49: `NonEmptyBalances` -> `NonEmptyWalletBalances`  ?

veado (approved):

Issue #650, opened 2020-12-02
[Ledger] Add business logic to send / sign BTC transactions


veado:
Add everything we do need to send / sign Bitcoin transactions by using Ledger. Currently we don't have UI prepared for it, but needed parts are:

- Prepare messaging `main` -> `renderer` thread
- Prepare service `pushLedgerTx$`

Part of #611

PR #661, opened 2020-12-08
[Ledger] Add business logic to send / sign BTC transactions

sarawut11:
This PR is for https://github.com/thorchain/asgardex-electron/issues/650


src/shared/const.ts
```diff
@@ -14,6 +14,9 @@ export enum ExternalUrl {
 export const LEDGER = {
   // Ledger now uses bip44 derivation path
   // For more details, please check - https://ledger.readthedocs.io/en/latest/background/hd_use_cases.html
-  GET_BTC_MAINNET_ADDRESS: "173'/0'/0'/0/0",
-  GET_BTC_TESTNET_ADDRESS: "173'/1'/0'/0/0"
+  BTC_DERIVE_PATH: {
+    MAINNET: "84'/0'/0'/0/0",
+    TESTNET: "84'/1'/0'/0/0"
+  },
```
veado, line 20: Make [derive_path](https://github.com/xchainjs/xchainjs-lib/blob/38645063168663fb23b364d48ae7698efa1ab37c/packages/xchain-bitcoin/src/client.ts#L58-L61) of xchain-btc public to import it here. See https://github.com/xchainjs/xchainjs-lib/issues/134


@sarawut11 Does it make sense to do it similar with `BNB_DERIVE_PATH_ARRAY` (extracting into `xchain-binance` etc.) ?

src/main/api/ledger.ts
```diff
@@ -78,3 +75,45 @@ export const getLedgerAddress = (chain: Chain, network: Network) => {
       break
   }
 }
+
+const signBtcTxInLedger = async (network: Network, ledgerTx: LedgerTxInfo) => {
```
veado, line 79: Type for return value: `Promise<E.Either...>>`

src/main/api/ledger.ts
```diff
@@ -78,3 +75,45 @@ export const getLedgerAddress = (chain: Chain, network: Network) => {
...
+    })
+    await transport.close()
+
+    return E.right(txHex)
+  } catch (error) {
+    return E.left(getErrorId(error.message, error.statusText))
+  }
+}
+
+export const signTxInLedger = (chain: Chain, network: Network, ledgerTx: LedgerTxInfo) => {
```
veado, line 110: Type for return value: `Promise<Either...>>`

src/renderer/services/bitcoin/ledger.ts
```diff
@@ -18,10 +22,53 @@ const retrieveLedgerAddress = (network: Network) =>
...
 
-const createLedgerService = (): LedgerService => ({
+export const createLedgerService = (): LedgerService => ({
   ledgerAddress$,
   retrieveLedgerAddress,
   removeLedgerAddress: () => setLedgerAddressRD(RD.initial)
 })
 
-export { createLedgerService }
+export const createLedgerTransferService = <Client extends XChainClient, T extends XChainClient$<Client>>(
```
veado, line 31: No need to create  another factory to have a `LedgerTransferService`, which seems very similar to `createTransferService` (see next comment).

src/renderer/services/bitcoin/ledger.ts
```diff
@@ -18,10 +22,53 @@ const retrieveLedgerAddress = (network: Network) =>
...
+        (e): TxLD =>
+          Rx.of(
+            RD.failure({
+              msg: e.toString(),
+              errorId: ErrorId.SEND_TX
+            })
+          )
+      ),
+      RxOp.startWith(RD.pending)
+    )
```
veado, line 65: No need to duplicate code of `tx$` of `src/renderer/services/clients/transaction/transfer.ts` here.

Try to create a stream using existing `tx$()` function and use it to update state with `setLedgerTxRD`. If you think it's easier to do it in `transfer.ts`, just go for it. It's still fine to have `ledgerXYZTransfer` functions in `transfer.ts`.

veado (changes requested):
Some thoughts about current architecture ...

src/renderer/services/bitcoin/ledger.ts
sarawut11: sec, this is not completed code, let's discuss about this in slack after I update this code

src/renderer/services/bitcoin/ledger.ts
```diff
@@ -18,10 +20,47 @@ const retrieveLedgerAddress = (network: Network) =>
...
+
+  const pushLedgerTx = (params: LedgerTxInfoParams): Rx.Subscription => ledgerTx$(params).subscribe(setLedgerTxRD)
+
+  return {
+    ledgerTxRD$,
+    pushLedgerTx,
+    resetLedgerTx: () => setLedgerTxRD(RD.initial)
+  }
+}
```
veado, line 66: Pls move everything of `createLedgerTransferService` into `createLedgerService`. Update `LedgerService` and remove `createLedgerTransferService`

veado (changes requested):
Looks good! Last change re

src/renderer/services/bitcoin/ledger.ts
```diff
@@ -20,47 +20,42 @@ const retrieveLedgerAddress = (network: Network) =>
...
+    switchMap((signResult) =>
+      Either.isRight(signResult)
+        ? Rx.from(broadcastTx({ txHex: signResult.right, nodeUrl: params.nodeUrl, nodeApiKey: params.nodeApiKey }))
+        : Rx.from(() =>
+            Promise.reject({
+              ledgerErrorId: signResult.left
             })
           )
-      )
+    ),
```
veado: BTW: You can use `E.fold` to avoid `Either.isRight(..) ? .. : ..`
```
    switchMap((signResult) =>
      FP.pipe(
        signResult,
        E.fold(
          (ledgerErrorId) =>
            Rx.from(
              Promise.reject({
                ledgerErrorId
              })
            ),
          (txHex) => Rx.from(broadcastTx({ txHex, nodeUrl: params.nodeUrl, nodeApiKey: params.nodeApiKey }))
        )
      )
    ),
```

veado (approved):

PR #662, opened 2020-12-08
[Withdraw] Update fee handling for multichain + Rune Native

### LINKED ISSUE #644: [Withdraw] Update fee handling for multichain + Rune Native
as described in #640

veado:
**Withdraw fees are based on following fees:**
- (1) `thorMemo` - Fee for sending memo transaction
- (2) `assetOut` - Outbounding fee on asset chain
- (3) `thorOut` - Outbounding fee on Thorchain

All three fees are stored into a single object (type `WithdrawFees`). 

**Validation are done as follow:**
- (1) `thorMemo` - Fee has to be covered by balance in wallet
- (2) `assetOut` - Outbounding fee has to be covered by expected amount to withdraw 
- (3) `thorOut` - Outbounding fee has to be covered by expected amount to withdraw 

**Pricing for  (2) `assetOut` fee is always priced in asset:**
- Example 1: Withdraw `BNB.BUSD` -> Price in `BUSD`
- Example 2: Withdraw `BTC.BTC` -> Price in `BTC`

**Changes:**
- [x] Update storybook addons, configuration, stories declarations (with help of @thatStrangeGuyThorchain )
- [x] Witdraw: Update styles
- [x] Witdraw: Update stories
- [x] Witdraw: Update fee handling
- [x] Deposit: Some renaming, remove `Helper.formatFee`
- [x] Deprecate `getRunePricePool` / ' getDefaultRunePricePool` to use `RUNE_PRICE_POOL` only
- [x] Update i18n
- [x] Update `depositFees$` to use `selectedPoolChain$` (but not `selectedPoolAsset$`)
Closes #644 
- [x] Services: Update `withdrawFees$` / `withdrawFeeByChain$`, Update `getWithdrawMemo$`, Update `selectedPricePool$`, Add ` WithdrawFees`, ` WithdrawFeesRD`, ` WithdrawFeesLD`
- [x] Remove `AssetRuneB1A` in favour of `AssetRuneNative` on different places (mostly in tests)
- [x] Remove usage of `BASE_CHAIN_ASSET` in `AddDepositView`


src/renderer/components/deposit/withdraw/Withdraw.tsx
```diff
@@ -1,65 +1,282 @@
...
+  fees: WithdrawFeesRD
 }
 
 export const Withdraw: React.FC<Props> = ({
+  type,
+  asset,
+  assetPoolData,
   onWithdraw,
-  depositAsset,
+  asset: depositAsset,
```
thatStrangeGuyThorchain, line 70: looks like duplicated

src/renderer/components/deposit/withdraw/Withdraw.tsx
```diff
@@ -1,65 +1,282 @@
...
-    [fee]
+    )
+  }, [oFees, oRuneBalance, withdrawPercent])
+
+  const renderThorMemoFeeError = useMemo(() => {
+    const runeBalance = FP.pipe(
+      oRuneBalance,
+      O.getOrElse(() => ZERO_BASE_AMOUNT),
+      baseToAsset
+    )
```
thatStrangeGuyThorchain, line 112: i know this is really small difference here but if we change order we can avoid calculations in case of `O.none`|

```ts
const runeBalance = FP.pipe(
      oRuneBalance,
      O.map(baseToAsset),
      O.getOrElse(() => ZERO_ASSET_AMOUNT),
    )
```

real nice feature of ADT is lazy functions. in more global sense it's better to use all advantages of this features

src/renderer/components/deposit/withdraw/Withdraw.tsx
```diff
@@ -1,65 +1,282 @@
...
+        // Missing (or loading) fees does not mean we can't sent something. No error then.
+        () => !O.isNone(oFees),
+        ([{ thorMemo }, balance]) => balance.amount().isLessThan(thorMemo.amount())
       )
-    ],
-    [fee]
+    )
+  }, [oFees, oRuneBalance, withdrawPercent])
+
+  const renderThorMemoFeeError = useMemo(() => {
```
thatStrangeGuyThorchain, line 107: maybe here is a sense to add check 
```ts
if (isThorMemoFeeError) {
  return null
}
```

why do we need all of that calculations if there is no error ?

src/renderer/components/deposit/withdraw/Withdraw.tsx
```diff
@@ -1,65 +1,282 @@
...
+      O.fold(
+        // Missing (or loading) fees does not mean we can't sent something. No error then.
+        () => !O.isNone(oFees),
+        (fee) => assetToBase(withdrawAmounts.assetWithdraw).amount().isLessThan(fee.amount())
+      )
+    )
+  }, [oAssetChainFee, oFees, withdrawAmounts.assetWithdraw, withdrawPercent])
+
+  const renderAssetChainFeeError = useMemo(() => {
+    return FP.pipe(
```
thatStrangeGuyThorchain, line 175: same here. maybe 
```ts
if(isAssetChainFeeError) {return null}
```

src/renderer/components/deposit/withdraw/Withdraw.tsx
```diff
@@ -1,65 +1,282 @@
...
+              {intl.formatMessage(
+                { id: 'deposit.withdraw.fees' },
+                { thorMemo: formattedThorMemoFee, assetOut: formattedAssetOutFee, thorOut: formattedThorOutFee }
+              )}
+            </>
+          )
+        }
+      )
+    )
+  }, [asset, calculateAssetChainFee, fees, intl, isDesktopView])
```
thatStrangeGuyThorchain, line 277: as we want to keep things simplier why don't we use `Fees` component ? detailed information we can place behind `i` popup icon. we can put this "additional info"  as a part of `Fees` component to have the same things everywhere

src/renderer/components/deposit/withdraw/Withdraw.tsx
```diff
@@ -105,26 +322,43 @@ export const Withdraw: React.FC<Props> = ({
...
+                {renderFee}
+              </Styled.FeeLabel>
+            </Col>
+          </Styled.FeeRow>
+          <Styled.FeeErrorRow>
+            <Col>
+              <>
+                {!!isThorMemoFeeError && renderThorMemoFeeError}
+                {!!isThorOutFeeError && renderThorOutFeeError}
+                {!!isAssetChainFeeError && renderAssetChainFeeError}
```
thatStrangeGuyThorchain, line 351: with propose checks for render methods we will not any conditions at the render itself

src/renderer/services/chain/fees/withdraw.ts
```diff
@@ -1,67 +1,65 @@
...
-      O.map(() => withdrawFeeByChain$(BASE_CHAIN)),
-      O.getOrElse((): LiveData<Error, BaseAmount> => Rx.of(RD.initial))
+const withdrawFees$: WithdrawFeesLD = Rx.combineLatest([selectedPoolChain$, reloadWithdrawFees$]).pipe(
+  RxOp.switchMap(([oPoolChain, _]) =>
+    FP.pipe(
+      oPoolChain,
+      O.map((chain) =>
+        FP.pipe(
+          Rx.combineLatest([withdrawFeeByChain$(chain), withdrawFeeByChain$('THOR')]),
+          RxOp.map(sequenceTRDFromArray),
```
thatStrangeGuyThorchain, line 52: looks like one liner
```ts
liveData.sequenceArray([withdrawFeeByChain$(chain), withdrawFeeByChain$('THOR')]),
```
or
```ts
liveData.sequenceT(withdrawFeeByChain$(chain), withdrawFeeByChain$('THOR')),
```

thatStrangeGuyThorchain (changes requested):

src/renderer/components/deposit/withdraw/Withdraw.tsx
veado: We have [a check in line 349](https://github.com/thorchain/asgardex-electron/pull/662/files/d13ce1e91288667eed01337ff61c1bdf12edb5c3#diff-9523733ca0dcea33995c77fb2a65e28b89a9873ae0e9443005f3180c08fc2f5eR349)

src/renderer/components/deposit/withdraw/Withdraw.tsx
veado: We have a [check in line 351](https://github.com/thorchain/asgardex-electron/pull/662/files/d13ce1e91288667eed01337ff61c1bdf12edb5c3#diff-9523733ca0dcea33995c77fb2a65e28b89a9873ae0e9443005f3180c08fc2f5eR349)

src/renderer/components/deposit/withdraw/Withdraw.tsx
veado: We can go with `Fees` component later if all implementations of fees will be done and we know what we really need in this component.

src/renderer/components/deposit/withdraw/Withdraw.tsx
veado: That's correct. However, it's just another way for doing same thing.

thatStrangeGuyThorchain: @veado not exactly. with your implementation all of that stuff will be computed anyway but it will not be shown.

src/renderer/components/deposit/withdraw/Withdraw.tsx
thatStrangeGuyThorchain: that's totally correct. maybe it's better to keep things simplier until we 100% understand what we want form the UI ? 
Fees component allows to gather all logic in a single place with a single logic. in case we need custom fees rendering we can implement it later. what do you think ?

veado: We already have a custom case here by providing different labels for desktop / mobile view. We don't have that in `Fees` (yet) and we should wait if it make sense for other parts, too. Let's just wait few days, then we start another round of `Fees` component (if needed).

src/renderer/components/deposit/withdraw/Withdraw.tsx
veado: You are right in case of computation - I will change it ...

veado: 8c4fad5

src/renderer/components/deposit/withdraw/Withdraw.tsx
veado: Fee labels for WITHDRAW:

**Desktop view**
Screenshot from 2020-12-14 19-48-49

**Mobile view**

Screenshot from 2020-12-14 19-49-17

thatStrangeGuyThorchain: and we still can have "mobile view" for all devices and (i) button with additional description about fees if semeone wants to see detailed info

Issue #640, opened 2020-12-01
[epic] Update fee handling for multichain + Rune Native


veado:
# Multichain fees

1) GasFee - this is what the user pays to send their asset, to someone else, or to THORChain.
2) `OutboundTransactionFee` is what the user is charged (in the asset) to pay for each outbound from THORChain (swap/removeLiquidity), it is basically 3times the `fast` gas cost for the destination chain

Both of these can be provided by the XChain clients. (1) is just what the user chooses, but (2) is always 3x the `fast` fee

## Swap #641

Fees | Needed | API
------------ | ------------- | -------------
Source Chain: sourceAsset | yes  | `getFees` (`xchain-source`) ~~will be what the user chooses~~ `fastest` as default
Destination Chain: destinationAsset | yes | `getFees` (`xchain-dest`) will be 3x the `fast` fee 

#### Example 1: Native Rune -> BTC

Fees | API
------------ | -------------
Source Chain: RUNE | `getFees`(`xchain-source`)  ~~will be what the user chooses~~ `fastest` as default
Destination Chain: BTC | `getFees` (`xchain-bitcoin`) this will be 3x the `fast` fee

## Deposit (symmetrical) #642

Fees | Needed | API
------------ | ------------- | -------------
THORChain (**) | yes | `getFees` (xchain-thorchain)
_Asset_ Chain (**) | yes | `getFees` (xchain client)

#### Example: sym. Deposit BTC

Fees | API
------------ | -------------
x Rune | `getFees` (`xchain-thorchain`)  ~~will be what the user chooses~~ `fastest` as default
x BTC | `getFees` (`xchain-bitcoin`)  ~~will be what the user chooses~~ `fastest` as default

## Deposit (asymmetrical) #643

Fees | Needed | API
------------ | ------------- | -------------
_Asset_ Chain (**) | yes | `getFees` (xchain client)  ~~will be what the user chooses~~ `fastest` as default

### Example: asym. Deposit BTC

Fees | API
------------ | -------------
x BTC | `getFees` (`xchain-bitcoin`)  ~~will be what the user chooses~~ `fastest` as default

## Withdraw #644

Fees | Needed | API
------------ | ------------- | -------------
x Rune | yes | `getFees` (`xchain-thorchain`)  ~~will be what the user chooses~~ `fastest` (as default)  to send the memo tx
_Asset_ Chain (**) | yes |  (`xchain-dest`) this will be 3x the `fast` fee
[...]

PR #663, opened 2020-12-10
[Swap] Update fee handling for multichain + Rune Native

### LINKED ISSUE #569: Swap: Add business logic to load fees

### LINKED ISSUE #641: [Swap] Update fee handling for multichain + Rune Native
as described in #640

thatStrangeGuyThorchain:
closes #641 
Closes #569


src/renderer/components/swap/Swap.styles.tsx
```diff
@@ -4,6 +4,8 @@ import { palette } from 'styled-theme'
 
 import { media } from '../../helpers/styleHelper'
 import { AssetInput as AssetInputBase } from '../uielements/assets/assetInput'
+import { Drag as BaseDrag } from '../uielements/drag'
+import { Label as UILabel } from '../uielements/label'
```
veado, line 8: [nitpick] Please use same name convention for elements imported from `uielements/`. In other `*.styles.tsx` files you will find something like `import { Comp as UIComp } from '../uielements/comp''

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +397,142 @@ export const Swap = ({
     )
   }, [assetsToSwap, onConfirmSwap, changeAmount, closePrivateModal])
 
+  const chainFees = useMemo(() => [sourceChainFee, targetChainFee].filter((fee) => !RD.isFailure(fee)), [
+    sourceChainFee,
+    targetChainFee
+  ])
+
+  // Need to collect all fees based on chain to compare it to final balance
+  const oFeesByChainHashMap = useMemo(
```
veado, line 406: Type `oFeesByChainHashMap`

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +397,142 @@ export const Swap = ({
     )
   }, [assetsToSwap, onConfirmSwap, changeAmount, closePrivateModal])
 
+  const chainFees = useMemo(() => [sourceChainFee, targetChainFee].filter((fee) => !RD.isFailure(fee)), [
```
veado, line 400: Type `chainFees`

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +397,142 @@ export const Swap = ({
...
+            acc[keyByAsset] = baseAmount(previousFee.plus(cur.amount.amount()))
+            return acc
+          })
+        ),
+        RD.toOption
+      ),
+    [chainFees]
+  )
+
+  const isDifferentCains = useMemo(
```
veado, line 423: Type `isDifferentCains`

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +397,142 @@ export const Swap = ({
...
+    () =>
+      FP.pipe(
+        sequenceTOption(sourceAsset, targetAsset),
+        O.chain(([source, target]) => (source.chain !== target.chain ? O.some(true) : O.none))
+      ),
+    [sourceAsset, targetAsset]
+  )
+
+  const sourceChainBalanceError = useMemo(
```
veado, line 432: Type `sourceChainBalanceError`

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +397,142 @@ export const Swap = ({
...
+            O.map((result) => result.isNegative()),
+            O.getOrElse(() => false)
+          )
+        }),
+        O.getOrElse(() => false)
+      ),
+    [changeAmount, sourceAsset, walletBalances, oFeesByChainHashMap]
+  )
+
+  const sourceChainErrorLabel = useMemo(() => {
```
veado, line 469: Type `sourceChainErrorLabel`

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +397,142 @@ export const Swap = ({
...
+    )
+
+    return (
+      <Styled.BalanceErrorLabel>
+        {intl.formatMessage({ id: 'swap.errors.amount.balanceShouldCoverChainFee' }, { asset: sourceChainAsset })}{' '}
+      </Styled.BalanceErrorLabel>
+    )
+  }, [sourceChainBalanceError, sourceAsset, intl])
+
+  const targetChainBalanceError = useMemo(
```
veado, line 487: Type `targetChainBalanceError`

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +397,142 @@ export const Swap = ({
...
+              return targetBalance.amount.amount().minus(feeAmount).isNegative()
+            })
+          )
+        }),
+        O.getOrElse(() => false)
+      ),
+    [targetAsset, isDifferentCains, walletBalances, oFeesByChainHashMap]
+  )
+
+  const targetChainErrorLabel = useMemo(() => {
```
veado, line 509: Type `targetChainErrorLabel`

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +397,142 @@ export const Swap = ({
...
+    )
+
+    return (
+      <Styled.ErrorLabel>
+        {intl.formatMessage({ id: 'swap.errors.amount.balanceShouldCoverChainFee' }, { asset: targetChainAsset })}{' '}
+      </Styled.ErrorLabel>
+    )
+  }, [targetChainBalanceError, targetAsset, intl])
+
+  const fees = useMemo(() => chainFees.map(RD.map((fee) => ({ asset: fee.chainAsset, amount: fee.amount }))), [
```
veado, line 527: Type `fees`

veado (changes requested):
Have you seen how we deal with fees for `Deposit`? Why not going a similar way? _I'm not sure if we really need all that logic in `Swap`, but let have a call first to understand things right ..._

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +397,142 @@ export const Swap = ({
...
+        // In this case we aclready included it in sourceChainFee
+        sequenceTOption(targetAsset, walletBalances, oFeesByChainHashMap, isDifferentCains),
+        O.chain(([targetAsset, walletBalances, feesByChainHashMap]) => {
+          return FP.pipe(
+            walletBalances,
+            A.findFirst((walletBalance) => eqAsset.equals(walletBalance.asset, targetAsset)),
+            O.map((targetBalance) => {
+              const targetChainAsset = getChainAsset(targetAsset.chain)
+              const feeAmount = feesByChainHashMap[assetToString(targetChainAsset)]?.amount() || ZERO_BN
+              return targetBalance.amount.amount().minus(feeAmount).isNegative()
```
veado, line 500: > OutboundTransactionFee is what the user is charged (in the asset)

^ https://github.com/thorchain/asgardex-electron/issues/640

An user won't be charged for target chain fees from its balances - it's charged in the target asset. To validate fee for target chain, we have to make sure that the expected amount of target is more than 3x target chain fee (see #640.

src/renderer/components/swap/Swap.tsx
```diff
@@ -51,12 +57,15 @@ type SwapProps = {
   poolDetails?: PoolDetails
   walletBalances?: O.Option<NonEmptyWalletBalances>
   nativeTxFee: NativeFeeRD
+  targetChainFee?: ChainFeeRD
+  sourceChainFee?: ChainFeeRD
```
veado, line 61: Provide a single property `fees` (similar to `AddDeposit`)

```
export type SwapFees = { source: BaseAmount; target: BaseAmount }
export type SwapFeesRD = RD.RemoteData<Error, SwapFees>
```
^ Similar to [`DepositFeesRD`](https://github.com/thorchain/asgardex-electron/blob/426654b801b689740ec326fc031a815cc9c10ed9/src/renderer/services/chain/types.ts#L34-L35)

With that ^, more logic can be moved from view into service layer.

src/renderer/services/chain/types.ts
```diff
@@ -17,13 +17,18 @@ export type FeeLD = LiveData<Error, BaseAmount>
...
+  source: SwapFeeRD
+  target: SwapFeeRD
 }
 
-export type ChainFeeRD = RD.RemoteData<Error, ChainFee>
-export type ChainFeeLD = LiveData<Error, ChainFee>
+export type SwapFeesLDS = {
+  source: SwapFeeLD
+  target: SwapFeeLD
+}
```
veado, line 31: ^ With suggested changes above, it's not needed anymore.

src/renderer/views/swap/SwapView.tsx
```diff
@@ -129,41 +125,39 @@ export const SwapView: React.FC<Props> = (_): JSX.Element => {
...
 
-  const targetChainFee = useObservableState(sourceChainFee$)
-  const sourceChainFee = useObservableState(targetChainFee$)
+  const swapFeesRDS: SwapFeesRDS = useMemo(
+    () => ({
+      source: sourceFee,
+      target: targetFee
+    }),
+    [sourceFee, targetFee]
+  )
```
veado, line 160: ^ With suggested changes before, no need to create `targetFee: SwapFeeRD` / `sourceFee: SwapFeeRD` in this view anymore.

src/renderer/services/chain/fees/swap.ts
```diff
@@ -0,0 +1,62 @@
...
+    case BNBChain:
+      return BNB.fees$
+
+    case THORChain:
+      return THOR.fees$
+
+    case BTCChain:
+      return FP.pipe(
+        BTC.fees$,
+        liveData.map((btcFees) => btcFees.fees)
```
veado, line 36: `BTC.fees$` is wrong! Check[ its documentation](https://github.com/thorchain/asgardex-electron/blob/20b3098c39748f4b96435856339fa652ee555d56/src/renderer/services/bitcoin/fees.ts#L35): `Transaction fees (no memo included)`. In case of `source` fee, a memo has to be passed to [`poolFee$`](https://github.com/thorchain/asgardex-electron/blob/20b3098c39748f4b96435856339fa652ee555d56/src/renderer/services/bitcoin/fees.ts#L71). Again, pls check implementation of `Deposit` or `Withdraw`!

src/renderer/services/chain/fees/swap.ts
```diff
@@ -0,0 +1,62 @@
...
+  BNB.reloadFees()
+  BTC.reloadFees()
+  THOR.reloadFees()
+}
+
+/**
+ * @todo rethink about using common fees at all views
+ * and move view-related calculations directly to the views
+ * instead of creating fee-services for every situations
+ */
```
veado, line 24: As discussed yesterday, we won't have common fees in case of BTC. Since it needs to know a memo to calculate fees and a memo will always be different. Also we don't want have much calculations in a view, which makes hard to test ...

veado (changes requested):

src/renderer/services/chain/types.ts
```diff
@@ -17,13 +17,18 @@ export type FeeLD = LiveData<Error, BaseAmount>
 export type FeesRD = RD.RemoteData<Error, Fees>
 export type FeesLD = LiveData<Error, Fees>
 
-export type ChainFee = {
-  chainAsset: Asset
-  amount: BaseAmount
+export type SwapFeeRD = RD.RemoteData<Error, BaseAmount>
+export type SwapFeeLD = LiveData<Error, BaseAmount>
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: yep. that was the first try. but i've moved them separately to separate fee-data.
edge case: swap between `BNB` and `BTC`. `BTC` fees can not be calculated with an empty client (without a phrase) and it means we will get an error for anything. i left separate  for the sake of users which have not added/unlock their walletes

src/renderer/components/swap/Swap.tsx
```diff
@@ -397,39 +387,31 @@ export const Swap = ({
...
   }, [assetsToSwap, onConfirmSwap, changeAmount, closePrivateModal])
 
-  const chainFees = useMemo(() => [sourceChainFee, targetChainFee].filter((fee) => !RD.isFailure(fee)), [
-    sourceChainFee,
-    targetChainFee
-  ])
-
-  // Need to collect all fees based on chain to compare it to final balance
-  const oFeesByChainHashMap = useMemo(
+  const sourceChainFee: RD.RemoteData<Error, Fee> = useMemo(
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: we know it here but we pass `[sourceChainFee, targetChainFee]` to the `Fees` component which still should know about an asset to format that values to the string in a right way

src/renderer/services/chain/fees/swap.ts
thatStrangeGuyThorchain: does it make sense to do it here ? previously i've checked swap's memo format and there is no amount information there. will this memo affect anything ?

src/renderer/components/swap/Swap.tsx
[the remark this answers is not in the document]
veado: If `Fees` component does need this information, you can pass it into. Still no need to have to use `Fee` type in `Swap` at other places I think...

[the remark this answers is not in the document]
thatStrangeGuyThorchain: this `Fee` is a fee prop type from `Fees` component. maybe we should rethink that type name

src/renderer/services/chain/fees/swap.ts
veado: If we send any memo with BTC, it will be added to OP_RETURN field of BTC tx. That's why fees will be calculated in a different way, check out [`calcFee` in xchain-bitcoin](https://github.com/xchainjs/xchainjs-lib/blob/ad73f5746196a55ff6a5967fe5f7af7524a5832d/packages/xchain-bitcoin/src/client.ts#L222-L229) to see the difference.

src/renderer/services/chain/types.ts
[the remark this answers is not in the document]
veado: We should always make sure that we have access to a client. If an user as unlocked its wallet or not a wallet imported (that's no client available), we can just hide the fee field.

src/renderer/services/chain/fees/swap.ts
veado: It seems you have already implemented it with 74175d0, right?

src/renderer/services/chain/types.ts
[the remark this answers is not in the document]
thatStrangeGuyThorchain: in this case we will not show ANY fee at all for user without a wallet/unlocked wallet. do we need this ?

src/renderer/services/binance/fees.ts
```diff
@@ -16,18 +17,21 @@ const { stream$: reloadFees$, trigger: reloadFees } = triggerStream()
  * Transaction fees
  */
 const fees$: (client: Client$) => FeesLD = (client$) =>
-  Rx.combineLatest([reloadFees$, client$]).pipe(
+  Rx.combineLatest([reloadFees$, client$.pipe(startWith(O.none as O.Option<Client>))]).pipe(
```
veado, line 20: Why do we need `startWith(O.none as O.Option<Client>)` here? Is not `O.none` the default value of any `Client$`?

src/renderer/services/chain/fees/swap.ts
thatStrangeGuyThorchain: @veado yes. for some reason i forgot to push that changes oO

src/renderer/services/binance/fees.ts
thatStrangeGuyThorchain: nice catch. seems this is an artifact from debugging

src/renderer/components/swap/Swap.tsx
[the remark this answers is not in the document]
thatStrangeGuyThorchain: it was built as array of RDs instead of RD of array for the sake of processing each fee separately. if one fee loading fails we still can show other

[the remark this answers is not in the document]
veado: To much details we don't need atm. If we got all fees, show all fees.

src/renderer/services/bitcoin/fees.ts
```diff
@@ -38,6 +38,7 @@ export const createFeesService = (oClient$: Client$): FeesService => {
     mergeMap(([oClient, _]) =>
       FP.pipe(
         oClient,
+        O.alt(() => O.some(new BitcoinClient({}))),
```
veado, line 41: Passing `nodeUrl` + `nodeApiKey` is needed here ^. Check [getFeesWithRates](https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-bitcoin/src/client.ts#L235) for details

src/renderer/services/chain/types.ts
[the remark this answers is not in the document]
veado: We decided to show still fees in case of unlocked / not imported wallet.

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +389,156 @@ export const Swap = ({
...
+      targetChainFee,
+      RD.getOrElse(() => ZERO_BASE_AMOUNT)
+    )
+
+    return FP.pipe(
+      targetAsset,
+      O.map((asset) => {
+        const chainAsset = getChainAsset(asset.chain)
+        return getValueOfAsset1InAsset2(chainFee, poolData[assetToString(chainAsset)], poolData[assetToString(asset)])
+      }),
```
veado, line 475: Another note:  Whenever asset === asset.chain you can't use `getValueOfAsset1InAsset2`. In this case just use [chainFee](https://github.com/thorchain/asgardex-electron/pull/663/files#diff-35dddf5143f357d65494359d2b087faeca6e8eb13706b5a652f73f5dfe3a86eeR465) - see [calculateAssetChainFee in Withdraw](https://github.com/thorchain/asgardex-electron/pull/662/files#diff-9523733ca0dcea33995c77fb2a65e28b89a9873ae0e9443005f3180c08fc2f5eR140-R146)

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +389,156 @@ export const Swap = ({
...
+        RD.map(([sourceFee, targetFee, sourceAsset, targetAsset]) => [
+          { asset: getChainAsset(sourceAsset.chain), amount: sourceFee },
+          { asset: targetAsset, amount: targetFee }
+        ])
+      ),
+    [sourceChainFee, targetChainFeeAmountInTargetAsset, sourceAsset, targetAsset]
+  )
+
+  const isSwapDisabled: boolean = useMemo(
+    () => changeAmount.eq(0) || FP.pipe(walletBalances, O.isNone) || sourceChainBalanceError || targetChainFeeError,
```
veado, line 526: Don't disable swap in case of `sourceChainBalanceError` or `targetChainFeeError`

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +389,156 @@ export const Swap = ({
...
+  )
+
+  const outputLabel = useMemo(
+    () =>
+      FP.pipe(
+        targetAsset,
+        O.map((asset) =>
+          formatAssetAmountCurrency({ amount: assetAmount(swapData.swapResult), asset, trimZeros: true })
+        ),
+        O.getOrElse(() => formatBN(swapData.swapResult, 7))
```
veado, line 537: Why is decimal `7` ?

src/renderer/i18n/en/swap.ts
```diff
@@ -9,7 +9,9 @@ const swap: SwapMessages = {
   'swap.balance': 'Balance',
   'swap.output': 'Output',
   'swap.drag': 'Drag to swap',
-  'swap.searchAsset': 'Search Asset'
+  'swap.searchAsset': 'Search Asset',
+  'swap.errors.amount.balanceShouldCoverChainFee': 'Balance for {asset} should cover a fee',
+  'swap.errors.amount.outputShouldCoverChainFee': 'Output should cover a fee {fee}'
```
veado, line 14: Could you use [similar wording as we will have in Withdraw](https://github.com/thorchain/asgardex-electron/pull/662/files#diff-db167fd729cf28a10cb5727884e1a15f833e3e662f901f6f4e63aa691145b4f6R33-R36) _(you could not know it before...)_ It will help users to see different fees.

src/renderer/services/chain/types.ts
```diff
@@ -17,6 +17,22 @@ export type FeeLD = LiveData<Error, BaseAmount>
 export type FeesRD = RD.RemoteData<Error, Fees>
 export type FeesLD = LiveData<Error, Fees>
 
+export type SwapFeeRD = RD.RemoteData<Error, BaseAmount>
+export type SwapFeeLD = LiveData<Error, BaseAmount>
+
+export type SwapFeesLDS = {
+  source: SwapFeeLD
+  target: SwapFeeLD
+}
```
veado, line 26: Why do you still need this ^ type? Use `SwapFeesLD` everywhere, also to push it into `Swap` (instead of using nested RemoteData in `SwapFeesLDS`).

src/renderer/components/swap/Swap.tsx
```diff
@@ -50,13 +56,14 @@ type SwapProps = {
...
   walletBalances?: O.Option<NonEmptyWalletBalances>
-  nativeTxFee: NativeFeeRD
   txWithState?: TxWithStateRD
   resetTx?: () => void
   goToTransaction?: (txHash: string) => void
   runeAsset: Asset
   activePricePool: PricePool
   PasswordConfirmation: React.FC<{ onSuccess: () => void; onClose: () => void }>
+  reloadFees?: () => void
+  fees?: SwapFeesRD
```
veado, line 66: Why not just `SwapFeesLD`?

veado (changes requested):

src/renderer/services/bitcoin/fees.ts
veado: Previous comment seems to be still relevant ^.

src/renderer/services/chain/types.ts
thatStrangeGuyThorchain: arstifact

src/renderer/components/swap/Swap.tsx
thatStrangeGuyThorchain: did not get you. 
'cuz this is `RD` not `LD` ? 🤔

src/renderer/components/swap/Swap.tsx
thatStrangeGuyThorchain: this was copy-pasted from prev implementation
https://github.com/thorchain/asgardex-electron/pull/663/files#diff-35dddf5143f357d65494359d2b087faeca6e8eb13706b5a652f73f5dfe3a86eeL440

src/renderer/components/swap/Swap.tsx
```diff
@@ -390,6 +389,156 @@ export const Swap = ({
...
+    const chainFee = FP.pipe(
+      targetChainFee,
+      RD.getOrElse(() => ZERO_BASE_AMOUNT)
+    )
+
+    return FP.pipe(
+      targetAsset,
+      O.map((asset) => {
+        const chainAsset = getChainAsset(asset.chain)
+        return getValueOfAsset1InAsset2(chainFee, poolData[assetToString(chainAsset)], poolData[assetToString(asset)])
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: yep. my bad. should check for `undefined`
but im gonna to disagree with using `getPoolDetail`. it uses `Array.prototype.find` method. it means we we need go through ALL items everytime we want to use it. this is why here was used a hash map

src/renderer/components/swap/Swap.tsx
veado: Yep, `SwapFeesRD` is fine - my bad - sorry!

src/renderer/components/swap/Swap.tsx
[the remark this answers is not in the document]
veado: > it uses Array.prototype.find method.

It's too early to worry about such performance things ... It can be done in 2022.

src/renderer/components/swap/Swap.tsx
veado: If we don't have an asset (it's still needed to swap anything), then we should just return ZERO_BN I think ...

src/renderer/services/bitcoin/fees.ts
thatStrangeGuyThorchain: seems no. this comment is outdated

src/renderer/components/swap/Swap.tsx
thatStrangeGuyThorchain: i suppose `swapData.swapResult` should contain this data. in case it calculated anything we should show it i believe

src/renderer/components/swap/Swap.tsx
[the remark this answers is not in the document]
thatStrangeGuyThorchain: > It's too early to worry about such performance things ... 

this is really dangerous question 😄 when is not it 'early' ? when we will see it by our eyes ? it will be too late IMO 😄 

hm? 
> will avoid to write extra code

`getPoolDetail` will return `O.Option`.  we will need to combine 2 options here and after use `toPoolData`.
seems this is the same as compare 2 values with `undefined` here but without handling each element in an array

src/renderer/components/swap/Swap.tsx
veado: Please just fix it (as [you suggested](https://github.com/thorchain/asgardex-electron/pull/663#discussion_r542635609) or [me before](https://github.com/thorchain/asgardex-electron/pull/663#discussion_r542619425))

thatStrangeGuyThorchain: @veado changed by [this commit](https://github.com/thorchain/asgardex-electron/pull/663/commits/68b7df576a586d70289e73f68944b115bdcb0caa)

veado (approved):
:tada:

PR #666, opened 2020-12-15
update dependencies, update codes according to the updated dependencies

sarawut11:


src/renderer/services/chain/fees/withdraw.ts
```diff
@@ -34,10 +34,10 @@ const withdrawFeeByChain$ = (chain: Chain): FeeLD => {
           )
         )
       )
-    case 'ETH':
-      return Rx.of(RD.failure(new Error('Deposit fee for ETH has not been implemented')))
     case 'THOR':
       return THOR.fees$.pipe(liveData.map(({ fast }) => fast))
+    default:
+      return Rx.of(RD.failure(Error(`Deposit fee for ${chain.toUpperCase} has not been implemented`)))
```
veado, line 40: Same here: Provide `failure` for all possible chains with proper error messages.

src/renderer/services/chain/fees/deposit.ts
```diff
@@ -94,10 +94,10 @@ const depositFeeByChain$ = (chain: Chain, type: DepositType): FeeLD => {
           )
         )
       )
-    case 'ETH':
-      return Rx.of(RD.failure(new Error('Deposit fee for ETH has not been implemented')))
```
veado, line 98: Pls don't remove it ^. Instead of, we should do same `failure` for chains we don't have implemented yet.

src/renderer/services/chain/fees/swap.ts
```diff
@@ -44,7 +35,7 @@ const feesByChain$ = (chain: Chain, memo?: Memo): FeesLD => {
         memo ? BTC.memoFees$(memo) : BTC.fees$,
         liveData.map((btcFees) => btcFees.fees)
       )
-    case ETHChain:
+    default:
       return Rx.of(RD.failure(Error(`${chain.toUpperCase} fees is not implemented yet`)))
```
veado, line 39: Pls add `failure` for all possible chains with a proper error message.

veado (changes requested):
Some thoughts about `case` handling of possible `Chain`'s.

veado (approved):

Issue #660, opened 2020-12-08
[RuneNative] Add upgrade feature


veado:
To upgrade `BNB.RUNE` -> `RUNE.RUNE`, provide a UI similar to send screen. 

- All amount of BNB.RUNE will be send
- An user **can't** add any amount or change an amount
- memo `SWITCH:thoraddress` (`thoraddress` -> address of ASGARD vault -> get it from /thorchain/inbound_addresses)

PR #669, opened 2020-12-16
Fix address handling + some refactor of balance handling

veado:
Few updates for balance handling:

- [x] Fix / improve `address` handling by putting it into an `Option` in case of failures
- [x] Make `balances$` failing in case of unaccessible client address 
- [x] Add `WalletType` to `ChainBalance` and `WalletBalance`
- [x] Make `walletAddress` in `ChainBalance` an `Option`
- [x] Some renamings on types, properties etc.

Needed for #660


src/renderer/contexts/WalletContext.tsx
```diff
@@ -37,7 +37,7 @@ const initialContext: WalletContextValue = {
   reloadBalances$,
   loadTxs,
   balancesState$,
-  chainBalances$,
+  chainBalances$: chainBalances$,
```
sarawut11, line 40: just chainBalances$

src/renderer/services/ethereum/balances.ts
```diff
@@ -4,34 +4,49 @@ import { AssetETH, baseAmount } from '@xchainjs/xchain-util'
...
+        Rx.from(client.getBalance()).pipe(
+          mergeMap((balance) =>
+            Rx.of(
+              RD.success({
+                asset: AssetETH,
+                amount: baseAmount(balance.toString(), ETH_DECIMAL),
+                walletAddress
+              })
+            )
+          ),
```
thatStrangeGuyThorchain, line 41: not sure but maybe this will be more readable ?
```ts
O.map((walletAddress) =>
    FP.pipe(
      Rx.of(client.getBalance()),
      RxOp.map((balance) =>
        RD.success({
          asset: AssetETH,
          amount: baseAmount(balance.toString(), ETH_DECIMAL),
          walletAddress
        })
      )
    )
  ),
  RxOp.switchMap(
    O.getOrElse(() =>
      Rx.of(
        RD.failure<ApiError>({
          errorId: ErrorId.GET_BALANCES,
          // TODO (@Veado) Add i18n
          msg: 'Cant get address from client'
        })
      )
    )
  )
```

src/renderer/services/wallet/types.ts
```diff
@@ -36,13 +38,25 @@ export type KeystoreService = {
   validatePassword$: (password: string) => LiveData<Error, null>
 }
 
+/**
+ * Wraps WalletBalancesRD into an object to provide extra information (`Address` + `Chain` + `WalletType`)
+ * Currently needed in `AssetView` - TODO(@Veado) Think about to extract it into view layer (as helper or so)
```
thatStrangeGuyThorchain, line 43: seems this is semantically correct way to store this data together as every operation is also depends on `WalletType`. is not it ?

thatStrangeGuyThorchain (dismissed):

src/renderer/services/wallet/types.ts
veado: Not sure if I understand your comment ^ correctly: `WalletBalance` (as part of `balances: WalletBalancesRD` in `ChainBalances`) already includes information about `Address` + `Chain` + `WalletType`. With that we have same data in `ChanBalances` - which is a duplication of data. However, as mentioned in the comment we use `ChainBalances`  in `AssetView` to build a nested table of balances. Each nested table has a header, which displays an `Address` and in near future it will show a wallet type as well. To show these data instantly in the header without the need to load balances, we have to put it into `ChainBalance` as extra properties besides `balances.`

src/renderer/services/ethereum/balances.ts
veado: Yep, it looks better. For now we can ignore it here, because `xchain-ethereum` will be updated soon and with this update `loadBalances$` will be gone here.

PR #673, opened 2020-12-18
[Upgrade feature] State handling (incl. few other fixes)

### LINKED ISSUE #660: [RuneNative] Add upgrade feature

veado:
- [x] State handling `TxLD` in `AssetDetailView`
- [x] Rename `pushTx` -> `subscribeTx` / `sendDepositTx` -> `sendTx`
- [x] Fix `getWalletBalanceByAsset` + fix its tests
- [x] Add test for `isRuneNativeAsset`

Closes #660

### DIFF
--- a/src/renderer/components/wallet/assets/AssetDetails.tsx
+++ b/src/renderer/components/wallet/assets/AssetDetails.tsx
@@ -1,12 +1,15 @@
-import React, { useCallback, useMemo, useState } from 'react'
+import React, { useCallback, useEffect, useMemo, useState } from 'react'
 
+import * as RD from '@devexperts/remote-data-ts'
 import { Address } from '@xchainjs/xchain-client'
-import { Asset, assetToString, BaseAmount } from '@xchainjs/xchain-util'
+import { Asset, AssetRuneNative, assetToString, BaseAmount } from '@xchainjs/xchain-util'
 import { Row, Col, Grid } from 'antd'
+import Modal from 'antd/lib/modal/Modal'
 import * as FP from 'fp-ts/lib/function'
 import * as O from 'fp-ts/lib/Option'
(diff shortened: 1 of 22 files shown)


src/renderer/components/wallet/assets/AssetDetails.tsx
```diff
@@ -53,6 +56,22 @@ export const AssetDetails: React.FC<Props> = (props): JSX.Element => {
...
+  // unsubscribe of possible previous subscription of upgrade tx
+  // It will be called whenever state of `bnbTxSub` changed
+  useEffect(() => {
+    return () => {
+      FP.pipe(
+        bnbTxSub,
+        O.map((sub) => sub.unsubscribe())
+      )
+    }
+  }, [bnbTxSub])
```
thatStrangeGuyThorchain, line 73: actually i still can not understand why you have decided to handle subscriptions at the View layer. 
why can not we use a subscriber-function from a service? (you have renamed it to the `subscribeTx`)

veado: The state of sending an upgrade tx is needed in `AssetDetails` only (to display it in tx modal). We don't need to store it "globally", because it's not be used anywhere else nor it's updated in a service by any other service (e.g. websocket service etc.) in this case. That was the only reason we have used this state globally before. 

A simple state handling in a component is just we do need here. If you have a case to store it globally, you can still use `subscribeTx`.

PR #684, opened 2020-12-22
add test for getBalanceByAsset, minor code update

### LINKED ISSUE #518: Test `getBalanceByAsset`
located in `src/renderer/services/wallet/util.ts`

sarawut11:
closes #518

### DIFF
--- a/src/renderer/services/wallet/util.test.ts
+++ b/src/renderer/services/wallet/util.test.ts
@@ -124,7 +124,7 @@ describe('services/wallet/util/', () => {
           amount: baseAmount(2)
         },
         {
-          asset: ASSETS_TESTNET.RUNE,
+          asset: AssetRuneNative,
           amount: baseAmount(0)
         }
       ] as WalletBalances
@@ -150,14 +150,14 @@ describe('services/wallet/util/', () => {
               asset: ASSETS_TESTNET.BNB
             },
             {
-              asset: ASSETS_TESTNET.RUNE
+              asset: AssetRuneNative
             }
           ] as WalletBalances,
           [AssetBTC.ticker, AssetETH.ticker, AssetRune67C.ticker, AssetBNB.ticker]
         )
       ).toEqual([
         {
-          asset: ASSETS_TESTNET.RUNE
+          asset: AssetRuneNative
         },
         {
           asset: ASSETS_TESTNET.BNB
--- a/src/renderer/services/wallet/util.test.ts
+++ b/src/renderer/services/wallet/util.test.ts
@@ -1,4 +1,4 @@
-import { AssetBNB, AssetBTC, AssetETH, AssetRune67C, baseAmount } from '@xchainjs/xchain-util'
+import { AssetBNB, AssetBTC, AssetETH, AssetRune67C, AssetRuneNative, baseAmount } from '@xchainjs/xchain-util'
 import { some, none } from 'fp-ts/lib/Option'
 
 import { ASSETS_TESTNET } from '../../../shared/mock/assets'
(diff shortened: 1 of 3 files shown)


src/renderer/services/wallet/util.ts
```diff
@@ -12,12 +10,12 @@ import { WalletBalance } from '../../types/wallet'
...
 
-export const getKeystoreContent = (state: KeystoreState): Option<KeystoreContent> => pipe(state, O.chain(identity))
+export const getKeystoreContent = (state: KeystoreState): Option<KeystoreContent> => FP.pipe(state, chain(FP.identity))
 
 export const getPhrase = (state: KeystoreState): Option<Phrase> =>
-  pipe(
+  FP.pipe(
     getKeystoreContent(state),
-    O.map(({ phrase }) => phrase)
+    map(({ phrase }) => phrase)
```
veado, line 18: Pls keep `O` as prefix. It will help us to distinguish different `map` from other modules / libraries. Same with `chain` etc.

Here are some naming conventions (I can share my VSCode snippets to you):
```
import * as A from 'fp-ts/Array'
import * as O from 'fp-ts/lib/Option'
import * as E from 'fp-ts/lib/Either'
... // and others
```

veado (changes requested):

src/renderer/services/wallet/util.test.ts
```diff
@@ -173,4 +174,32 @@ describe('services/wallet/util/', () => {
...
+            asset: ASSETS_TESTNET.BOLT
+          },
+          {
+            asset: ASSETS_TESTNET.FTM
+          },
+          {
+            asset: ASSETS_TESTNET.BNB
+          },
+          {
+            asset: ASSETS_TESTNET.RUNE
```
veado, line 195: Pls don't use `ASSETS_TESTNET.RUNE` anymore - it will be deprecated in the very near future. Use `AssetRuneNative`.

thatStrangeGuyThorchain (approved):

src/renderer/services/wallet/util.test.ts
```diff
@@ -173,4 +174,32 @@ describe('services/wallet/util/', () => {
...
+          {
+            asset: ASSETS_TESTNET.FTM
+          },
+          {
+            asset: ASSETS_TESTNET.BNB
+          },
+          {
+            asset: AssetRuneNative
+          }
+        ] as WalletBalances)
```
veado, line 197: As discussed in DM: 
- Don't use `as` for type casting, it's just like cheating the type system. 
- Use `eqOAsset` to test result in a very safety way

thatStrangeGuyThorchain: @veado IMHO its ok to use typecasting in tests. sometimes we could need only 1 or 2 values for unit testing but target interface might be really big. and if something changes what does not affect tests we will have tests failed and need to fix it. anyway if there is something missed we will find it out when tests fail

veado: @thatStrangeGuyThorchain Don't agree. It's just the lack of TypeScript you can do it like that. And it's the knowledge you might have about internals of a function like `getBalanceByAsset`. However, if something of its internal will be changed, this way won't make sure, that tests will fail then. Create mock factories if you scare about building "bigger" objects for tests. Or change types of function parameters and return types to be more generic. But that's not what we do need in our code in this case, where we have to make sure we will get a `WalletBalance` (not parts) from it.

sarawut11: yeah, make sense

veado (approved):

PR #686, opened 2020-12-23
Update pricing to use RuneNative

### LINKED ISSUE #683: Update pricing to use RuneNative
to remove `RuneB1A` and `AssetRune67C` from any price calculation.

In ASGDX an user can do a price selection by using a dropdown in the header (see attached screen shot). Currently prices are based on `BNB.RUNE`, but for multichain we have to switch prices to RuneNative (`THOR.RUNE`). 

First part of this task is to understand how pricing works. For doing that, check current implementation in the code (starting by `HeaderPriceSelector.tsx` <- that's the component where the user can change a "price asset" and checking all the business logic behind up to services). After that, change everything needed for using `THOR.RUNE` as the only "Rune" price asset. 

Screenshot from 2020-12-22 17-46-00

Part of #575

sarawut11:
closes #683

### DIFF
--- a/src/renderer/components/deposit/add/AddDeposit.stories.tsx
+++ b/src/renderer/components/deposit/add/AddDeposit.stories.tsx
@@ -2,7 +2,7 @@ import React from 'react'
 
 import * as RD from '@devexperts/remote-data-ts'
 import { storiesOf } from '@storybook/react'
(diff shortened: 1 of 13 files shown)


veado (approved):
LGTM

PR #691, opened 2020-12-27
[SWAP] fix transaction data information #688

### LINKED ISSUE #688: [SWAP] fix transaction data information
currently we have getting of pool datas for multichain txs. it leads to wrong results showing to user
image

image

expected: results should be similar

thatStrangeGuyThorchain:
closes #688

as we have now simple pricing for source and target assets it's not clear for user what this data is. this pr includes changing simple showing of assets prices to swap results based on selected price asset 

Including few tweaks:
- Remove box shadow from `transparent` button
- Remove `calResult` in favour of `slip`property
- Remove `visible` property
- Update stories for using latest storybook
- Remove `swap.state.pending` from i18n

### DIFF
--- a/src/renderer/components/swap/Swap.tsx
+++ b/src/renderer/components/swap/Swap.tsx
@@ -39,7 +39,6 @@ import { TxStatus, TxTypes } from '../../types/asgardex'
 import { PricePool } from '../../views/pools/Pools.types'
 import { CurrencyInfo } from '../currency'
 import { SwapModal } from '../modal/swap'
-import { CalcResult } from '../modal/swap/SwapModal.types'
 import { AssetSelect } from '../uielements/assets/assetSelect'
 import { Fee, Fees } from '../uielements/fees'
 import { Modal } from '../uielements/modal'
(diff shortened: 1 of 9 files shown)


src/renderer/components/swap/Swap.tsx
```diff
@@ -316,43 +316,48 @@ export const Swap = ({
...
+                    assetToBase(assetAmount(changeAmount)),
+                    poolData[assetToString(sourceAssetWP.asset)],
+                    activePricePool.poolData
+                  )
+
+                return (
+                  <SwapModal
+                    key={'swap modal result'}
+                    basePriceAsset={activePricePool.asset}
+                    calcResult={{ slip: swapData.slip } as CalcResult}
```
veado, line 341: Avoid such unsafe casting, since `{ slip: swapData.slip }` is not `CalcResult`! If `SwapModal` does need `{ slip: swapData.slip}` only, change its property `calcResult to `slip: BigNumber`.

veado (changes requested):

src/renderer/components/swap/Swap.tsx
veado: Fix 505655a8a41e82159652a7a834516f568ed117ba

veado (approved):

Issue #725, opened 2021-01-11
[AddDeposit] Add extra content to TxModel


veado:
depending on `sym` / `asym` deposit

PR #730, opened 2021-01-11
Fix styles: SwapAssets, StepBar, Trend

veado:
+ update stories

Thx to @thatStrangeGuyThorchain for pointing out about missing `className` ([needed by `styled components`](https://styled-components.com/docs/basics#styling-any-component))

As part of #725

### DIFF
--- a/src/renderer/components/uielements/assets/assetData/AssetData.tsx
+++ b/src/renderer/components/uielements/assets/assetData/AssetData.tsx
@@ -22,17 +22,27 @@ export type Props = {
   price?: BaseAmount
   priceAsset?: PricePoolAsset
   size?: Styled.AssetDataSize
+  // `className` is needed by `styled components`
+  className?: string
 }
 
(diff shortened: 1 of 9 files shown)


src/renderer/components/uielements/assets/assetData/AssetData.tsx
```diff
@@ -22,17 +22,27 @@ export type Props = {
...
 export const AssetData: React.FC<Props> = (props): JSX.Element => {
-  const { asset, amount: assetAmount, noTicker = false, price = baseAmount(0), priceAsset, size = 'small' } = props
+  const {
+    asset,
+    amount: assetAmount,
+    noTicker = false,
+    price = baseAmount(0),
+    priceAsset,
+    size = 'small',
+    className = ''
```
thatStrangeGuyThorchain, line 37: just a nitpick
usually `className` is optional and there is no need to use default value

thatStrangeGuyThorchain (approved):

PR #748, opened 2021-01-13
[Settings] Mini-Addresses

### LINKED ISSUE #739: [Settings] Mini-Addresses
If not enough room to show the full address, always show as prefix123...123, never any other format.

Provide a helper function to "minify" addresses we can use everywhere.

As part of #732

junkai121:
Fix #739

### DIFF
--- a/src/renderer/helpers/addressHelper.ts
+++ b/src/renderer/helpers/addressHelper.ts
@@ -1,4 +1,4 @@
-import { BNBChain, CosmosChain, THORChain } from '@xchainjs/xchain-util'
+import { BNBChain, THORChain } from '@xchainjs/xchain-util'
 
 export const truncateAddress = (addr: string, chain: string, network: string): string => {
   const first = addr.substr(0, getAddressPrefixLength(chain, network) + 3)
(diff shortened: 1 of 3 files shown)


src/renderer/helpers/addressHelper.ts
```diff
@@ -0,0 +1,20 @@
...
+export const truncateAddress = (addr: string, chain: string, network: string): string => {
+  const first = addr.substr(0, getAddressPrefixLength(chain, network) + 3)
+  const last = addr.substr(addr.length - 3, 3)
+  return `${first}...${last}`
+}
+
+export const getAddressPrefixLength = (chain: string, network: string): number => {
+  switch (chain) {
+    case BNBChain:
+      return network === 'testnet' ? 4 : 3
```
veado, line 12: Maybe `client.getPrefix` using xchain-binance ([code](https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-binance/src/client.ts#L200-L202)) ?

src/renderer/helpers/addressHelper.ts
```diff
@@ -0,0 +1,20 @@
...
+    case THORChain:
+      return network === 'testnet' ? 5 : 4
```
veado, line 14: Unfortunately `xchain-thorchain` [does not export `getPrefix` - it still private](https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-thorchain/src/client.ts#L232-L234). Let's change it (see https://github.com/xchainjs/xchainjs-lib/issues/183)

veado (approved):

PR #757, opened 2021-01-14
Add business logic for asym deposit txs

veado:
- [x] Business logic for asym deposit txs
- [x] Use latest `@thorchain/asgardex-util@0.8.0`

Part of #537


src/renderer/services/chain/transaction/deposit.ts
```diff
@@ -0,0 +1,130 @@
...
+                O.getOrElse(() => state)
+              ),
+            // ignore `failure` state + return same state (no changes)
+            () => state,
+            // ignore `success` state + return same state (no changes)
+            () => state
+          )
+        )
+      )
+    ),
```
thatStrangeGuyThorchain, line 127: in general seems here is no need to use combination of `switchMap -> Rx.of`, it looks like simple `map`

thatStrangeGuyThorchain (approved):
LGTM.
just one small nitpick

PR #779, opened 2021-01-20
fix styling issues after check for night/dark mode artifacts

sarawut11:
- fix caret position issue in wallet assetsview table
- fix text style issue under the upgrade button
- fix input component style issue


veado (approved):
LGTM

PR #784, opened 2021-01-21
UI Improvements

sarawut11:
- Use lowercase for addresses and transaction ID everywhere
- Use address truncation for all address/tx
- Add small copy button next to all address/txID

### DIFF
--- a/src/renderer/components/uielements/addressEllipsis/AddressEllipsis.tsx
+++ b/src/renderer/components/uielements/addressEllipsis/AddressEllipsis.tsx
@@ -1,30 +1,41 @@
 import { useCallback } from 'react'
 
+import { Address } from '@xchainjs/xchain-client'
 import { Chain } from '@xchainjs/xchain-util'
 
+import { Network } from '../../../../shared/api/types'
 import { truncateAddress } from '../../../helpers/addressHelper'
+import * as Styled from './AddressEllipsis.style'
+
+/**
+ * Custome address ellipsis component
+ * Based on https://github.com/bluepeter/react-middle-ellipsis/
+ */
 
 type Props = {
-  chain: string
-  network: string
+  address: Address
+  chain: Chain
+  network: Network
 }
 
 export const AddressEllipsis: React.FC<Props> = (props): JSX.Element => {
-  const { chain, network } = props
+  const { address, chain, network } = props
   const prepEllipse = useCallback(
-    (node: HTMLElement) => {
+    (node: HTMLElement, txtToEllipse: HTMLElement, copyIcon: HTMLElement) => {
       const parent = node.parentElement
       if (parent) {
-        const txtToEllipse: HTMLElement | null = node.childNodes[0] as HTMLElement
-        const copyIcon: HTMLElement | null = node.childNodes[1] as HTMLElement
-
-        if (txtToEllipse !== null) {
+        if (txtToEllipse) {
           if (txtToEllipse.hasAttribute('data-original')) {
             txtToEllipse.textContent = txtToEllipse.getAttribute('data-original')
           }
 
           if (
-            !isEllipse(node.offsetWidth > parent.offsetWidth ? parent : node, txtToEllipse, chain, network) &&
+            !ellipse({
+              parentNode: node.offsetWidth > parent.offsetWidth ? parent : node,
+              txtNode: txtToEllipse,
+              chain,
+              network
+            }) &&
             copyIcon
           ) {
             copyIcon.hidden = true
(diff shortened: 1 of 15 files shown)


src/renderer/components/wallet/txs/receive/Receive.tsx
```diff
@@ -79,15 +101,11 @@ export const Receive: React.FC<Props> = (props): JSX.Element => {
...
-                <Styled.Address size="large">{addressLabel}</Styled.Address>
-              </label>
-            ) : (
-              <Styled.AddressWrapper htmlFor="clipboard-btn">
-                <Styled.Address size="large">{addressLabel}</Styled.Address>
-              </Styled.AddressWrapper>
-            )}
+            <label htmlFor="clipboard-btn">
+              <Styled.Address size="large">
+                {isDesktopView ? addressLabel : truncateAddress(addressLabel, chain as Chain, network)}
```
veado, line 106:
```
chain as Chain
```
Don't cheat! If you don't have a network and chain, but you do need it to run `truncateAddress`, then you can't use `truncateAddress`. Or re-think about implementation of props for `Receive` component or of `truncateAddress` - but don't cheat ;)

src/renderer/components/wallet/txs/table/TxsTable.tsx
```diff
@@ -23,12 +25,31 @@ type Props = {
...
   const isDesktopView = Grid.useBreakpoint()?.lg ?? false
+  const network = useMemo(
+    () =>
+      FP.pipe(
+        oNetwork,
+        O.getOrElse(() => '')
+      ),
+    [oNetwork]
+  )
+  const chain = useMemo(
```
veado, line 44: What's type of `chain`? It seems like `Chain | string`, but you want to have `Chain` in `AddressEllipsis`, right?

veado (changes requested):
Re-think about usage of `O.Option` in general...

src/renderer/components/uielements/addressEllipsis/AddressEllipsis.tsx
```diff
@@ -39,29 +50,36 @@ export const AddressEllipsis: React.FC<Props> = (props): JSX.Element => {
...
+      const prepareEllipse = () => prepEllipse(node, node.childNodes[0], node.childNodes[1])
       if (node !== null) {
-        window.addEventListener('resize', () => {
-          prepEllipse(node)
-        })
-        prepEllipse(node)
+        window.addEventListener('resize', prepareEllipse)
+        prepEllipse(node, node.childNodes[0], node.childNodes[1])
+      } else {
+        window.removeEventListener('resize', prepareEllipse)
```
veado, line 58: Is  `removeEventListener` really called when `AddressEllipsis` will be unmounted? Something like follow should do such thing:

```
useEffect(() => {
    return () => window.removeEventListener('resize', prepareEllipse)
}, [])
```

src/renderer/components/wallet/txs/receive/Receive.tsx
```diff
@@ -103,7 +86,7 @@ export const Receive: React.FC<Props> = (props): JSX.Element => {
           <Styled.Div>
             <label htmlFor="clipboard-btn">
               <Styled.Address size="large">
-                {isDesktopView ? addressLabel : truncateAddress(addressLabel, chain as Chain, network)}
+                {!isDesktopView && chain ? truncateAddress(addressLabel, chain, network) : addressLabel}
```
veado, line 89: Instead of 
```
chain ? truncateAddress(addressLabel, chain, network) : addressLabel
```

you can provide a `renderAddress` function as follow:
```
const renderAddress = useMemo(() => {
  FP.pipe(
    oChain,
    O.fold(
      () => addressLabel,
      (chain) => truncateAddress(addressLabel, chain, network)
  )

}, [oChain])
```

src/renderer/components/wallet/txs/table/TxsTable.tsx
```diff
@@ -71,9 +54,9 @@ export const TxsTable: React.FC<Props> = (props): JSX.Element => {
     (address: Address, key: string) => (
       <Styled.Text key={key}>
         <div>
-          <AddressEllipsis chain={chain} network={network}>
-            <span>{address}</span>
-            <Styled.CopyLabel copyable={{ text: address }} />
+          <AddressEllipsis address={address} chain={chain} network={network}>
+            {/* <span>{address}</span>
+            <Styled.CopyLabel copyable={{ text: address }} /> */}
```
veado: ^ can be removed

veado (changes requested):

src/renderer/components/wallet/txs/table/TxsTable.tsx
veado: ^ done in 143698ba1c2b40011a0c702ab21a849267c528e7

veado (approved):
Code LGTM

Issue #774, opened 2021-01-19
[EPIC] Improvements from 19 Jan Call


thorchain-admin:
@sarawut11 
* Stats/Faq page use inspiration from BEPSwap (~~remove stats/faq button from footer at the meantime~~ **DONE** #777)
* Network page use inspiration from https://viewblock.io/thorchain (~~remove network button from footer at the meantime~~ **DONE** #777) 
* ~~Check for night/dark mode artifacts~~ **DONE** #779
* ~~Use *lowercase* for addresses and transaction ID everywhere~~ **DONE** #784
* ~~Use address truncation for all address/tx~~ **DONE** #784
* ~~Add small copy button next to all address/txID~~ **DONE** #784
* ~~Hide asset row if no asset for that chain~~ **DONE** #793 
* ~~Swap page inverse rate~~ **DONE** #789 
* ~~Import keystore: Errors messages have to be more explicit (e.g. password or keystore error)~~ **DONE** #777
* Bigger input fields (swap. deposit, withdraw)

@junkai121 
* Finalise Ethereum
* Add BitcoinCash (and testnet - google options for API)
* Add Litecoin (and testnet - sochain.com)

@thatStrangeGuyThorchain 
* ~~Finish Deposit screen~~ **DONE** #776 
* Add node Bond screen - let node operator add their node address, then show it all the details

@veado 
* Finish add/remove liquidity
* Show "Pool Shares" page to show which pools and how much

PR #793, opened 2021-01-23
hide asset row if no asset for that chain

sarawut11:
part of #774

### DIFF
--- a/src/renderer/components/wallet/assets/AssetsTableCollapsable.tsx
+++ b/src/renderer/components/wallet/assets/AssetsTableCollapsable.tsx
@@ -276,15 +276,12 @@ export const AssetsTableCollapsable: React.FC<Props> = (props): JSX.Element => {
   useEffect(() => {
     // don't change openPanelKeys if user has already changed panel state
     if (!collapseChangedByUser) {
+      // filter out empty list of balances
       const keys = FP.pipe(
         chainBalances,
         A.map(({ balances }) => balances),
         A.filterMap(RD.toOption),
-        A.filterMapWithIndex((i, balances) =>
-          balances.length === 0 || (balances.length === 1 && balances[0].amount.amount().isZero())
-            ? O.none
-            : O.some(i.toString())
-        )
+        A.filterMapWithIndex((i, balances) => (balances.length > 0 ? O.some(i.toString()) : O.none))
       )
       setOpenPanelKeys(keys)
     }
--- a/src/renderer/components/wallet/assets/AssetsTableCollapsable.tsx
+++ b/src/renderer/components/wallet/assets/AssetsTableCollapsable.tsx
@@ -275,7 +275,10 @@ export const AssetsTableCollapsable: React.FC<Props> = (props): JSX.Element => {
   useEffect(() => {
     // don't change openPanelKeys if user has already changed panel state
     if (!collapseChangedByUser) {
-      const keys = chainBalances.map((_, i) => i.toString())
+      const keys: string[] = []
+      chainBalances.forEach((chainBalance, index) => {
+        if (RD.isSuccess(chainBalance.balances) && chainBalance.balances.value.length) keys.push(index.toString())
+      })
       setOpenPanelKeys(keys)
     }
   }, [chainBalances, collapseChangedByUser])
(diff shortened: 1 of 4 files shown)


src/renderer/services/clients/balances.ts
```diff
@@ -44,6 +44,7 @@ const loadBalances$ = ({
       (walletAddress) =>
         Rx.from(client.getBalance(walletAddress)).pipe(
           map(RD.success),
+          liveData.map(A.filter((balance) => !balance.amount.amount().isZero())),
```
veado, line 47: Please note: We don't want to exclude zero balances everywhere in the ui. For example: If an user has sent all amount of an token using `SEND` view, we might want to show balances there (which are 0 now). 

If we do need a filter for zero balances in asset view, do such filter there (but not here in this service here  ^).

src/renderer/components/wallet/assets/AssetsTableCollapsable.tsx
```diff
@@ -275,7 +275,10 @@ export const AssetsTableCollapsable: React.FC<Props> = (props): JSX.Element => {
   useEffect(() => {
     // don't change openPanelKeys if user has already changed panel state
     if (!collapseChangedByUser) {
-      const keys = chainBalances.map((_, i) => i.toString())
+      const keys: string[] = []
+      chainBalances.forEach((chainBalance, index) => {
+        if (RD.isSuccess(chainBalance.balances) && chainBalance.balances.value.length) keys.push(index.toString())
+      })
```
veado, line 281: Don't access the "inside" value by using `chainBalance.balances.value`. You can do it like the following (check the comments, you can remove those later in code)

```
      const keys = FP.pipe(
        chainBalances,
        // get `balances: WalletBalancesRD` from `chainBalances`
        A.map(({ balances }) => balances),
        // Filter `success` balances and get its value (WalletBalance) from it
        A.filterMap(RD.toOption),
        // Filter empty list of balances and get a `key` from `i` (index)
        A.filterMapWithIndex((i, balances) => (balances.length > 0 ? O.some(i.toString()) : O.none))
      )
```

veado (changes requested):

src/renderer/components/wallet/assets/AssetsTableCollapsable.tsx
```diff
@@ -275,10 +276,16 @@ export const AssetsTableCollapsable: React.FC<Props> = (props): JSX.Element => {
...
-      const keys: string[] = []
-      chainBalances.forEach((chainBalance, index) => {
-        if (RD.isSuccess(chainBalance.balances) && chainBalance.balances.value.length) keys.push(index.toString())
-      })
+      const keys = FP.pipe(
+        chainBalances,
+        A.map(({ balances }) => balances),
+        A.filterMap(RD.toOption),
+        A.filterMapWithIndex((i, balances) =>
+          balances.length === 0 || (balances.length === 1 && balances[0].amount.amount().isZero())
```
veado, line 284:
```
balances.length === 0 || balances[0].amount.amount().isZero()
```

src/renderer/components/wallet/assets/AssetsTableCollapsable.tsx
```diff
@@ -275,10 +276,16 @@ export const AssetsTableCollapsable: React.FC<Props> = (props): JSX.Element => {
   useEffect(() => {
     // don't change openPanelKeys if user has already changed panel state
     if (!collapseChangedByUser) {
-      const keys: string[] = []
-      chainBalances.forEach((chainBalance, index) => {
-        if (RD.isSuccess(chainBalance.balances) && chainBalance.balances.value.length) keys.push(index.toString())
-      })
+      const keys = FP.pipe(
```
veado, line 279: Pls add a small note above `const keys`, something like `filter out empty list of balances and zero balances.

src/renderer/views/wallet/AssetsView.tsx
```diff
@@ -16,7 +18,10 @@ export const AssetsView: React.FC = (): JSX.Element => {
   const history = useHistory()
   const { chainBalances$ } = useWalletContext()
 
-  const chainBalances = useObservableState(chainBalances$, [] as ChainBalances)
+  const chainBalances = FP.pipe(
+    useObservableState(chainBalances$, [] as ChainBalances),
+    A.filter((chainBalance) => RD.isSuccess(chainBalance.balances))
+  )
```
veado, line 24: ^ That's not correct if we still want to show loading and error states of each chain.

veado (changes requested):

veado (approved):

PR #795, opened 2021-01-25
Provide StorageContext / StorageService

### LINKED ISSUE #127: Provide StorageContext / StorageService
to store user preferences, `basePriceAsset` etc. 

Link:
- Simple data persistence for your Electron app or module - Save and load user preferences, app state, cache, etc  https://github.com/sindresorhus/electron-store

thatStrangeGuyThorchain:
- [x] introduced `api/fileStore` service for interacting with files at the electron-part
- [x] exposed `commonStorage`  public api to the renderer part
- [x] introduced storage service at renderer part
- [x] introduced typesafe common approach to create file-based storages
- [x] refactored i18n context to take data form storage service

closes #127


src/main/api/fileStore.ts
```diff
@@ -0,0 +1,79 @@
...
+const APP_NAME = app?.name ?? 'ASGARDEX'
+
+const APP_DATA_DIR = path.join(app?.getPath('appData') ?? './testdata', APP_NAME)
+export const STORAGE_DIR = path.join(APP_DATA_DIR, 'storage')
+
+// If `fullFilePathname' does not exist, `fs.remove` silently does nothing.
+// @see https://github.com/jprichardson/node-fs-extra/blob/master/docs/remove.md
+const removeFile = async (fullFilePathname: string) => fs.remove(fullFilePathname)
+
+const isFileExist = async (fullFilePathname: string) => fs.pathExists(fullFilePathname)
```
veado, line 18: ^ Not the same, but very similar to [`removeKeystore` and `keystoreExist` ](https://github.com/thorchain/asgardex-electron/blob/e7beb5b78802c3b635c90ec8eb9eb1a76b3247e1/src/main/api/keystore.ts#L22-L24). Maybe we can bring it "together" by using `removeFile` / `isFileExists` in _keystore.ts_

src/main/api/fileStore.ts
```diff
@@ -0,0 +1,79 @@
...
+    return Promise.resolve({ ...defaultValue, ...fileContent })
+  } catch {
+    // Resolve with default value in case of any error
+    return Promise.resolve(defaultValue)
+  }
+}
+/**
+ * Replaces store-file's content with partial provided data
+ */
+const saveFile = async <T>(fullFilePathname: string, data: Partial<T>, defaultValue: T) => {
```
veado, line 40: `saveFile` might be misleading here, since it does not just save some data into a file, it append data to existing stuff. Maybe `appendDataToFile` or something ....?

src/main/api/fileStore.ts
```diff
@@ -0,0 +1,79 @@
...
+    const fileContent = await fs.readJSON(fullFilePathname)
+    // Combine file content with provided default value
+    return Promise.resolve({ ...defaultValue, ...fileContent })
+  } catch {
+    // Resolve with default value in case of any error
+    return Promise.resolve(defaultValue)
+  }
+}
+/**
+ * Replaces store-file's content with partial provided data
```
veado, line 38: [nitpick] ^ It does not replace, it appends new data to existing data in a file

src/main/preload.ts
```diff
@@ -10,6 +11,11 @@ import { apiUrl } from './api/url'
 // even if `contextIsolation` is enabled
 // https://www.electronjs.org/docs/api/context-bridge
 
+/**
+ * When exposing anything to the real world do not forget to
+ * declare appropriate types for global Window interface
+ * at the src/shared/api/types.ts
+ */
```
veado, line 18: :+1:

src/main/api/fileStore.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  try {
+    // Try to read and parse appropriate JSON store-file
+    const fileContent = await fs.readJSON(fullFilePathname)
+    // Combine file content with provided default value
+    return Promise.resolve({ ...defaultValue, ...fileContent })
+  } catch {
+    // Resolve with default value in case of any error
+    return Promise.resolve(defaultValue)
```
veado, line 34: What's the reason to never fail `getFileContent` by returning `defaultValue`? With it, `saveFile` will never fail as well, but in some cases we might want to know that we can't write something to disk, don't we?

veado (approved):
Pretty nice! Some questions / suggestions ...

src/main/api/fileStore.ts
thatStrangeGuyThorchain: hm. actually `appendDataToFile` is misleading too i believe. in my understanding append will work like:
old file content
`{"some": "data here}`
after append {newData: "new data here"} i assume content to be smth like this
`{"some": "data here}{"newData: "new data here""}`
as `append` method of the `fs` package

src/main/api/fileStore.ts
thatStrangeGuyThorchain: true. but i decided not to touch a code which is not supposed to be refactored at this issue for the sake of time and to keep PR smaller.  despite on the fact it's not a huge one it includes a lot of logical connections which should be kept in mind and changing previous implementation would increase that logic. lets try to generalize this things a little bit later

src/main/api/fileStore.ts
thatStrangeGuyThorchain: forgot to change it. i'd say `combines`. do you mind `combines` ?

src/main/api/fileStore.ts
thatStrangeGuyThorchain: for `getFileContent` i believe it makes sense to have default values for real world app.

about `saveFile` there is [fs.writeJSON](https://github.com/thorchain/asgardex-electron/pull/795/files#diff-1a3af8eeb2e0ceea21cd19c157aab914739e4ad090353a6710c61ab20917a93bR43) which will fail in case when it can not write to the file and all resulted `Promise` will be rejected. do i miss anything ?

src/main/api/fileStore.ts
veado: Makes sense.

src/main/api/fileStore.ts
veado: `saveToFile` :100:

src/main/api/fileStore.ts
veado: [Current implementation of `getFileContent`](https://github.com/thorchain/asgardex-electron/blob/0d27dbd376f39c287293a955d9b733d0910570a4/src/main/api/fileStore.ts#L16-L32) will never fail, there is no `try/catch` or something [around fs.writeJSON](https://github.com/thorchain/asgardex-electron/blob/0d27dbd376f39c287293a955d9b733d0910570a4/src/main/api/fileStore.ts#L19). We can do it, just let you know about ...

veado (dismissed):

veado (approved):

Issue #763, opened 2021-01-15
Generate API types / endpoints of Thornode API


veado:
based on https://thornode.ninerealms.com/thorchain/doc/swagger.json 

## TODO

Add to `package.json`

```json
"generate:types:thornode": "TS_POST_PROCESS_FILE=./node_modules/.bin/prettier openapi-generator-cli generate -i https://thornode.ninerealms.com/thorchain/doc/swagger.json -g typescript-rxjs -o ./src/renderer/types/generated/thornode --reserved-words-mappings in=in --enable-post-process-file ",
"clean:types:thornode": "rimraf ./src/renderer/types/generated/thornode",
```

## Current issue
[UPDATE April 13th 2022] [swagger.json](https://thornode.ninerealms.com/thorchain/doc/swagger.json) still does not fit requirements to generate types / endpoint from it. 

Example of missing types in generated code:
```ts

    /**
     * Retrieve all nodes that have bond
     * Get all nodes
     */
    getAllNodes = (): Observable<Array<object>> => {
        return this.request<Array<object>>({
            path: '/thorchain/nodes',
            method: 'GET',
        });
    };

```
^ `src/renderer/types/generated/thornode/apis/NodesApi.ts` (not checked in repository yet)

`object` should be a proper type, but it's not defined in [swagger.json](https://thornode.ninerealms.com/thorchain/doc/swagger.json)

PR #818, opened 2021-02-01
Generate API types / endpoints of Thornode API #763

### LINKED ISSUE #783: [Upgrade] Extend UI to enter an amount to upgrade
Before that ^, extract upgrade things from `AssetDetails` into it new `Upgrade` component. This `Upgrade` component will look very similar to `Send` component.

thatStrangeGuyThorchain:
added thornode api generation based on https://testnet.thornode.thorchain.info/thorchain/doc/swagger.json

known issue: while there is not fully valid spec we use `--skip-validate-spec` flag while generate api typings. by default not-found types for array will be fall-backed to string type and this is fine for now

closes #783


veado (approved):

veado:
We decided to add current needed (but not all) endpoints and types "by hand". _Closing for now ..._

Issue #507, opened 2020-10-06
Wallet: Add Pool Shares


veado:
All content goes into `wallet` -> `pool shares` (see [Asset View at Figma](https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020?node-id=1307%3A3)).

Add a table ([FIGMA layout](https://files.slack.com/files-pri/TBFG8JBBQ-F01LKL70664/image.png)) with all data of pool shares. Data are similar as we described in BEPSwap issue[ ADD: Pool Share in Wallet Pane #1121](https://gitlab.com/thorchain/bepswap/bepswap-web-ui/-/issues/1121)

![image](https://user-images.githubusercontent.com/61792675/106365321-0ef0a300-6335-11eb-9621-6fbebadd8a73.png)

PR #822, opened 2021-02-02
Add pool shares ui

sarawut11:
a part of #507


src/renderer/i18n/de/wallet.ts
```diff
@@ -3,6 +3,7 @@ import { WalletMessages } from '../types'
 const wallet: WalletMessages = {
   'wallet.nav.deposits': 'Deposits',
   'wallet.nav.bonds': 'Bonds',
+  'wallet.nav.poolshares': 'Pool Shares - DE',
```
veado, line 6: 'wallet.nav.poolshares': 'Poolanteile',

veado (changes requested):
Code looks good! Just some notes about i18n ...

veado (approved):

PR #833, opened 2021-02-04
Fix data for pending pools

veado:
This PR fixes missing pending pools in pools overview

### DIFF
--- a/src/renderer/services/midgard/pools.ts
+++ b/src/renderer/services/midgard/pools.ts
@@ -144,19 +144,38 @@ const createPoolsService = (
    * Ddata of pending `Pools` from Midgard
    */
   const apiGetPoolsPending$: PoolDetailsLD = apiGetPools$({ status: GetPoolsStatusEnum.Staged })
-  /**
-   * Data of all `Pools` or by given status
-   */
-  const apiGetPoolsByStatus$: (status?: GetPoolsStatusEnum) => PoolDetailsLD = (status) => apiGetPools$({ status })
 
+  /**
+   * Data of all `Pools`
+   */
+  const apiGetPoolsAll$: PoolDetailsLD = apiGetPools$({ status: undefined })
+
+  /**
+   * Helper to get (same) stream of `PoolDetailsLD` by given status
+   *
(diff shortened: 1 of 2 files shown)


src/renderer/services/midgard/pools.ts
```diff
@@ -145,6 +145,11 @@ const createPoolsService = (
    * Ddata of pending `Pools` from Midgard
    */
   const apiGetPoolsPending$: LiveData<Error, PoolDetails> = apiGetPools$({ status: GetPoolsStatusEnum.Staged })
+  /**
+   * Data of all `Pools` or by given status
+   */
+  const apiGetPoolsByStatus$: (status?: GetPoolsStatusEnum) => LiveData<Error, PoolDetails> = (status) =>
```
thatStrangeGuyThorchain, line 151: just nitpick
`PoolDetailsLD`

src/renderer/services/midgard/pools.ts
```diff
@@ -145,6 +145,11 @@ const createPoolsService = (
...
+    apiGetPools$({ status })
```
thatStrangeGuyThorchain, line 152: and just to clarify. why do we need this ? it only calls `apiGetPools$` factory with `{status}`. is not it ?

thatStrangeGuyThorchain (approved):
just a question. merge it if you wish

src/renderer/services/midgard/pools.ts
veado: If `status = undefined`, data of all pools (pending + availabe) are requested

thatStrangeGuyThorchain: ok

thatStrangeGuyThorchain (approved):
LGTM!

src/renderer/services/midgard/pools.ts
veado: [Last commit](https://github.com/thorchain/asgardex-electron/pull/833/commits/4119783bbe59d79085800b8d5e4a1e457fd24353) has improved it ^

PR #838, opened 2021-02-04
Disable `Eth` temporary

### LINKED ISSUE #816: Disable ETH temporary
For [next release](https://github.com/thorchain/asgardex-electron/milestone/1) we want to disable ETH temporary.

Try to find a "generic" solution, so that we can enable (or disable) chains based on `REACT_APP_CHAINS` defined in in `.env`. Strings of chains are based [on definition in xchain-util ](https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-util/src/chain.const.ts#L37)

```bash
`REACT_APP_CHAINS_ENABLED` = THOR,BNB,BTC
```

junkai121:
Fix #816

### DIFF
--- a/src/renderer/helpers/envHelper.ts
+++ b/src/renderer/helpers/envHelper.ts
@@ -11,3 +11,11 @@ export const isEnv = (env: ENV | undefined): env is ENV => !!env
  * @param defaultValue {string} Default value
  */
 export const envOrDefault = (env: ENV | undefined, defaultValue: string) => (isEnv(env) ? env : defaultValue)
+
+/**
+ * Read enabled chains from a given ENV.
+ * @param env {string} ENV
+ */
+export const loadEnabledChains = (env: ENV | undefined) => {
+  return envOrDefault(env, 'THOR,BNB,BTC,ETH').split(',')
+}
--- a/src/renderer/helpers/envHelper.ts
+++ b/src/renderer/helpers/envHelper.ts
@@ -11,11 +11,3 @@ export const isEnv = (env: ENV | undefined): env is ENV => !!env
  * @param defaultValue {string} Default value
  */
 export const envOrDefault = (env: ENV | undefined, defaultValue: string) => (isEnv(env) ? env : defaultValue)
-
-/**
- * Read enabled chains from a given ENV.
- * @param env {string} ENV
- */
-export const loadEnabledChains = (env: ENV | undefined) => {
-  return envOrDefault(env, 'THOR,BNB,BTC,ETH').split(',')
-}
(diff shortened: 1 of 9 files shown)


src/renderer/helpers/envHelper.ts
```diff
@@ -11,3 +11,11 @@ export const isEnv = (env: ENV | undefined): env is ENV => !!env
  * @param defaultValue {string} Default value
  */
 export const envOrDefault = (env: ENV | undefined, defaultValue: string) => (isEnv(env) ? env : defaultValue)
+
+/**
+ * Read enabled chains from a given ENV.
+ * @param env {string} ENV
+ */
+export const loadEnabledChains = (env: ENV | undefined) => {
+  return envOrDefault(env, 'THOR,BNB,BTC,ETH').split(',')
```
veado, line 20: I would suggest to use `THOR,BNB,BTC` as default, so we can use `REACT_APP_CHAINS_ENABLED` for development only. 

For production, `REACT_APP_CHAINS_ENABLED` don't need to be set. That's whenever we release a new chain, we have to update it here in code, but not in [ Github SECRETS of ASGDX](https://github.com/thorchain/asgardex-electron/settings/secrets/actions)

veado (changes requested):
Nice! Just one suggestions for handling default values ...

src/renderer/helpers/envHelper.ts
junkai121: Make sense

src/renderer/services/const.ts
```diff
@@ -1,9 +1,9 @@
 import { Network } from '../../shared/api/types'
-import { loadEnabledChains } from '../helpers/envHelper'
+import { envOrDefault } from '../helpers/envHelper'
 
 export const DEFAULT_NETWORK: Network = 'testnet'
 export const AVAILABLE_NETWORKS: Network[] = ['testnet', 'chaosnet', 'mainnet']
-export const ENABLED_CHAINS = loadEnabledChains(process.env.REACT_APP_CHAINS_ENABLED)
+export const ENABLED_CHAINS = envOrDefault(process.env.REACT_APP_CHAINS_ENABLED, 'THOR,BNB,BTC').split(',')
```
veado, line 6: Pls add `process.env.REACT_APP_CHAINS_ENABLED` to env.sample with a comment: 

```
Add chains you want to enable in development mode.
```

Also a similar note to README (in chapter [Development](https://github.com/thorchain/asgardex-electron#development)):

## How to enable (more) chains for development

Add chains you want to have in development to `REACT_APP_CHAINS_ENABLED`, for example

```bash
REACT_APP_CHAINS_ENABLED=THOR,BNB,BTC,ETH,LTC
```

veado (changes requested):

veado (approved):

PR #882, opened 2021-02-10
[swap] Enable swap calculation without wallet connected #875

### LINKED ISSUE #875: [swap] Enable swap calculation without wallet connected

thatStrangeGuyThorchain:
- added alt value in case there is no unlocked wallet

closes #875

### DIFF
--- a/src/renderer/components/swap/Swap.tsx
+++ b/src/renderer/components/swap/Swap.tsx
@@ -207,10 +207,13 @@ export const Swap = ({
               { ...maxAmount }
             : baseAmount(targetAmount.amount(), maxAmount.decimal)
         ),
+        // Provide alt value ONLY in case there is no unlocked wallet
+        // In other case missing oAssetWB means user does not have appropriate balance
+        O.alt(() => (isLocked(keystore) ? O.some(targetAmount) : O.none)),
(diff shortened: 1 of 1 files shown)


veado (approved):

src/renderer/components/swap/Swap.tsx
```diff
@@ -207,10 +207,13 @@ export const Swap = ({
               { ...maxAmount }
             : baseAmount(targetAmount.amount(), maxAmount.decimal)
         ),
+        // Provide alt value ONLY in case there is no unlocked wallet
+        // In other case missing oAssetWB means user does not have appropriate balance
+        O.alt(() => (isLocked(keystore) ? O.some(targetAmount) : O.none)),
```
veado, line 212: What about not imported wallet?

thatStrangeGuyThorchain: yep. `isLocked` also [check for hasImportedKeystore](https://github.com/thorchain/asgardex-electron/blob/master/src/renderer/services/wallet/util.ts#L26)

veado: Nevermind, `isLocked` already includes such a check

PR #885, opened 2021-02-11
enable avg slip and trades column in pool overview table

sarawut11:


src/renderer/services/midgard/pools.ts
```diff
@@ -104,13 +99,9 @@ const createPoolsService = (
                 FP.pipe(
                   // As midgard v2 is missing poolSlipAverage and swappingTxCount
                   // field we have to get them from getSwapHistory
+                  // Emulate ALL period of a pool life
```
veado, line 102: [question] Where do you emulate something? _Maybe this line ^ is just an artifact and can be removed...?_

veado (approved):
LGTM! Just a question...

src/renderer/services/midgard/pools.ts
sarawut11: yes it can be removed, so if we don't set `from`, `to` in the query, it assumes `all time` by default.

PR #922, opened 2021-02-16
Use wepback-node-externals

veado:
**DON'T MERGE** - for internal test only


veado:
It does not fix signing issue for macOS. _Closing..._

PR #949, opened 2021-02-19
[Settings] Add copy function for address in desktop view

### LINKED ISSUE #893: Release V0.0.4 - Please add copy function for address
image

Please add the a copy icon that copied to the clip board. I see that being used more than the need to go to a block explorer - happy to have both. 
BEPSwap has a copy function - good to keep it the same
image

junkai121:
Fix #893


src/renderer/components/wallet/settings/Settings.tsx
```diff
@@ -97,6 +97,22 @@ export const Settings: React.FC<Props> = (props): JSX.Element => {
     [oPhrase]
   )
 
+  const renderAddressWithBreak = useMemo(
+    () => (chain: Chain, address: Address, linkIcon: React.ReactElement) => (
```
veado, line 101: It looks like something for `useCallback`, isn't?

```ts
const renderAddressWithBreak = useCallback((chain: Chain, address: Address, linkIcon: React.ReactElement) => (...), [selectedNetwork])
```

veado (changes requested):

veado (approved):

PR #959, opened 2021-02-22
Resolve "[LTC] Support LTC for Swap"

### LINKED ISSUE #946: [LTC] Support LTC for Swap
as a part of #899

thatStrangeGuyThorchain:
- added swap fees for LTC
- added swap functionality for LTC

closes #946


src/renderer/services/chain/client.ts
```diff
@@ -21,6 +22,8 @@ export const clientByChain$ = (chain: Chain): XChainClient$ => {
       return ETH.client$
     case 'THOR':
       return THOR.client$
+    case 'LTC':
```
veado, line 25: [nitpick] LTCChain

src/renderer/services/chain/fees/fees.helper.ts
```diff
@@ -27,6 +28,9 @@ export const reloadDepositFeesByChain = (chain: Chain) => {
     case 'THOR':
       THOR.reloadFees()
       break
+    case 'LTC':
```
veado, line 31: [nitpick] LTCChain

src/renderer/services/litecoin/fees.ts
```diff
@@ -1,7 +1,51 @@
...
-  C.createFeesService
+export const createFeesService: ({ client$, chain }: { client$: Client$; chain: Chain }) => FeesService = ({
+  client$,
+  chain
+}) => {
+  const baseClient = C.createFeesService({ client$, chain })
+
+  /**
+   * Observable to load transaction fees
+   * If a client is not available, it returns an `initial` state
```
veado, line 22: [nitpick] Comment about client ^ seems to be deprecated...

veado (commented):
Code LGTM! Just few nitpick...

Pls don't merge before xchain-ltc has been updated for using other API then Sochain (similar to https://github.com/xchainjs/xchainjs-lib/pull/244).

src/renderer/helpers/fp/from.ts
```diff
@@ -0,0 +1,3 @@
+import * as O from 'fp-ts/Option'
+
+export const optionFromNullableString = O.fromPredicate<string | undefined | null, string>((s): s is string => !!s)
```
veado, line 3: [nitpick] `optionFromNullableString` -> `optionFromString`

src/renderer/services/litecoin/common.ts
```diff
@@ -26,7 +28,12 @@ const litecoinNetwork$: Observable<ClientNetwork> = network$.pipe(
   })
 )
 
-const SOCHAIN_URL = envOrDefault(process.env.REACT_APP_SOCHAIN_URL, 'https://sochain.com/api/v2')
+const LTC_NODE_TESTNET_URL = envOrDefault(process.env.REACT_APP_LTC_NODE_TESTNET_URL, '')
+const LTC_NODE_MAINNET_URL = envOrDefault(process.env.REACT_APP_LTC_NODE_MAINNET_URL, '')
+const oLTCNodePassword = optionFromNullableString(envOrDefault(process.env.REACT_APP_LTC_NODE_PASSWORD, ''))
+const oLTCNodeUsername = optionFromNullableString(envOrDefault(process.env.REACT_APP_LTC_NODE_USERNAME, ''))
```
veado, line 34: Please add default values using our proxy node - with it we don't need to add ENV to GH secrets and using ENV for development only.

src/renderer/services/litecoin/common.ts
```diff
@@ -26,7 +28,12 @@ const litecoinNetwork$: Observable<ClientNetwork> = network$.pipe(
...
+
+const ltcAuth = FP.pipe(sequenceSOption({ username: oLTCNodeUsername, password: oLTCNodePassword }), O.toUndefined)
```
veado, line 36: ^ It can be removed then (see previous comment)

veado (changes requested):

src/renderer/services/litecoin/common.ts
thatStrangeGuyThorchain: just to clarify why i've set this things exactly this way for some reasons:
for URLs:
- leave a possibility for user to define his own node url via env vars (in future. this is not implemented yet)
- empty string will be processed by [xchain-litecoin client by double negation of nodeUrl parameter](https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-litecoin/src/client.ts#L68) so `xchain-litecoin` will still use its default value for appropriate network

for auth info:
- the same possibility for future for user (node operator) to define his own node's auth info

if both username and password were provided by env this info will be passed to the `xchain-litecoin` client in other case there will be passed `undefined` value and [default auth info for litecoin client will be used](https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-litecoin/src/client.ts#L74)

so with this implementation NO NEED to add anything to ENV to GH secrets

what do you think about it ?

src/renderer/helpers/fp/from.ts
thatStrangeGuyThorchain: actually initially i set name to `optionFromString` but then i've added `nullable` to it 'cuz this function creates value from nullable values too. what do you think ?

src/renderer/services/litecoin/common.ts
veado: > empty string will be processed by xchain-litecoin client by double negation of nodeUrl parameter so xchain-litecoin will still use its default value for appropriate network

What happened if xchain-ltc will change it internal implementation?

Let's keep it simple and provide default values for now.

thatStrangeGuyThorchain: 👌

src/renderer/helpers/fp/from.ts
veado: Both names are valid. It was just a thought about this name, because we still have types to describe what a function expects as parameters. Anyway, I'm fine with both names :smile:

veado (approved):

PR #963, opened 2021-02-23
fix eth pricing in assets view

### LINKED ISSUE #961: [ETH] Wrong pricing for ETH in assets overview
:eyes: $ 93,839,104,811.967 :arrow_down: 

Screenshot from 2021-02-22 19-04-20

sarawut11:
closes #961

### DIFF
--- a/src/renderer/components/wallet/assets/AssetsTableCollapsable.tsx
+++ b/src/renderer/components/wallet/assets/AssetsTableCollapsable.tsx
@@ -13,7 +13,7 @@ import { useIntl } from 'react-intl'
 import { useHistory } from 'react-router-dom'
 
 import { Network } from '../../../../shared/api/types'
-import { isRuneBnbAsset } from '../../../helpers/assetHelper'
+import { ETH_DECIMAL, isEthAsset, isRuneBnbAsset } from '../../../helpers/assetHelper'
 import { getPoolPriceValue } from '../../../helpers/poolHelper'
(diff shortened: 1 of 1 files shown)


src/renderer/components/wallet/assets/AssetsTableCollapsable.tsx
```diff
@@ -145,7 +145,12 @@ export const AssetsTableCollapsable: React.FC<Props> = (props): JSX.Element => {
       const oPrice = getPoolPriceValue(balance, poolDetails, pricePool.poolData)
       const label = FP.pipe(
         oPrice,
-        O.map((price) => formatAssetAmountCurrency({ amount: baseToAsset(price), asset: pricePool.asset, decimal: 3 })),
+        O.map((price) => {
+          if (isEthAsset(balance.asset)) {
+            price.decimal = ETH_DECIMAL
+          }
```
veado, line 151: Pls keep in mind: If we have to do similar change for `price.decimal` at other places for ETH, extract that logic into a helper for pricing (into `getPoolPriceValue` in this case) - just to use same logic everywhere then ...

veado (approved):

src/renderer/components/wallet/assets/AssetsTableCollapsable.tsx
sarawut11: ok 👍

PR #903, opened 2021-02-14
[wip] Improve subscribing / unsubscribing of Rx.Subscriptions #898

thatStrangeGuyThorchain:
closes #898

PR #975, opened 2021-02-24
Introduce `useSubscriptionState`

### LINKED ISSUE #898: Improve subscribing / unsubscribing of Rx.Subscriptions
Currently we do some custom subscribing / unsubscribing Rx.Subscriptions instead of using `useSubscription` / `useObservableState` of [observable-hooks](https://observable-hooks.js.org/). 

For example in  [Upgrade.tsx](https://github.com/thorchain/asgardex-electron/blob/69685c486ce26a042cccbaaa6faebfb791101d68/src/renderer/components/wallet/txs/upgrade/Upgrade.tsx#L82-L106) :arrow_down: 

```
// (Possible) subscription of upgrade tx
  const [upgradeTxSub, _setUpgradeTxSub] = useState<O.Option<Rx.Subscription>>(O.none)
  // unsubscribe upgrade$ subscription
  const unsubscribeUpgradeTxSub = useCallback(() => {
    FP.pipe(
      upgradeTxSub,
      O.map((sub) => sub.unsubscribe())
    )
  }, [upgradeTxSub])
  const setUpgradeTxSub = useCallback(
    (state) => {
      unsubscribeUpgradeTxSub()
      _setUpgradeTxSub(state)
    },
    [unsubscribeUpgradeTxSub]
  )
  useEffect(() => {
    // Unsubscribe of (possible) previous subscription of `deposit$`
    return () => {
      unsubscribeUpgradeTxSub()
    }
  }, [unsubscribeUpgradeTxSub])
```

The only reason not to use `useSubscription` / `useObservableState` instead of a custom subscribing / unsubscribing is, that  `useSubscription` / `useObservableState` can't be use in an `useCallback` . But in some cases we do need it there (e.g. in Upgrade.tsx [here](https://github.com/thorchain/asgardex-electron/blob/69685c486ce26a042cccbaaa6faebfb791101d68/src/renderer/components/wallet/txs/upgrade/Upgrade.tsx#L182-L194))

veado:
based on first approach created by @thatStrangeGuyThorchain in #903

PR includes following changes:
- [x] Introduce `useSubscriptionState`
- [x] Remove `loadBalances` from SendViewXYZViews
- [x] Fix target of `Back` button to go back to asset details view (instead
of going back to form)

Closes #898

### DIFF
--- a/src/renderer/components/wallet/txs/send/SendFormBTC.tsx
+++ b/src/renderer/components/wallet/txs/send/SendFormBTC.tsx
@@ -305,6 +305,7 @@ export const SendFormBTC: React.FC<Props> = (props): JSX.Element => {
 
   return (
     <>
+      <div>amountToSend: {amountToSend.amount().toString()}</div>
       <Row>
         <Styled.Col span={24}>
           <AccountSelector
--- a/src/renderer/components/wallet/txs/send/SendFormBTC.tsx
+++ b/src/renderer/components/wallet/txs/send/SendFormBTC.tsx
@@ -305,7 +305,6 @@ export const SendFormBTC: React.FC<Props> = (props): JSX.Element => {
 
   return (
     <>
-      <div>amountToSend: {amountToSend.amount().toString()}</div>
       <Row>
         <Styled.Col span={24}>
           <AccountSelector
--- a/src/renderer/hooks/useSubscriptionState.ts
+++ b/src/renderer/hooks/useSubscriptionState.ts
@@ -30,15 +30,6 @@ export const useSubscriptionState = <T>(initialState: T) => {
     subRef.current = O.none
   }, [subRef])
 
-  // Store subscription by unsubscribing previous subscription
-  const setSub = useCallback(
-    (subscription) => {
-      unsubscribeSub()
-      subRef.current = O.some(subscription)
-    },
-    [unsubscribeSub]
-  )
-
   // Reset subscription and state
   const reset = useCallback(() => {
     unsubscribeSub()
@@ -47,8 +38,9 @@ export const useSubscriptionState = <T>(initialState: T) => {
 
   // Subscribe to an Observable
   const subscribe = (stream$: Rx.Observable<T>) => {
+    unsubscribeSub()
     const subscription = stream$.subscribe(setState)
-    setSub(subscription)
+    subRef.current = O.some(subscription)
   }
 
   /** Clean up */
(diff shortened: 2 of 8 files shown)


src/renderer/components/wallet/txs/send/SendFormBTC.tsx
```diff
@@ -305,6 +305,7 @@ export const SendFormBTC: React.FC<Props> = (props): JSX.Element => {
 
   return (
     <>
+      <div>amountToSend: {amountToSend.amount().toString()}</div>
```
veado, line 308: Remove it ^

src/renderer/hooks/useSubscriptionState.ts
```diff
@@ -0,0 +1,62 @@
...
+  // Reset subscription and state
+  const reset = useCallback(() => {
+    unsubscribeSub()
+    setState(initialState)
+  }, [unsubscribeSub, initialState])
+
+  // Subscribe to an Observable
+  const subscribe = (stream$: Rx.Observable<T>) => {
+    const subscription = stream$.subscribe(setState)
+    setSub(subscription)
```
thatStrangeGuyThorchain, line 51: i'd add unsibscribing prev subscription before creating a new one
```ts
const subscribe = (stream$: Rx.Observable<T>) => {
    unsubscribeSub() // <==== here
    const subscription = stream$.subscribe(setState)
   // ...
```
in general there might be really hard-to-detect edge cases caused by JS's event loop.  since subscribing at 50 line and setting sub at the 51 line there is really small time tick where 2 subscriptions might exist.
note: just a kind of paranoia you know :)

src/renderer/hooks/useSubscriptionState.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /** Clean up */
+  useEffect(() => {
+    return () => {
+      unsubscribeSub()
+    }
+  }, [unsubscribeSub])
+
+  return { state, setState, subscribe, reset }
```
thatStrangeGuyThorchain, line 61: do we need to provide internal `setState` as a part of `useSubscriptionState`'s result? it might be a little confusing i suppose and if someone would like to use `setState` it will lead to incosistency between stream's value and `state`'s value itself. i mean here should be the only source which can change inner state i believe

thatStrangeGuyThorchain (commented):
just few suggestions

PR #1006, opened 2021-03-01
Enable FR

veado:
+ remove duplicated type `Locale`

## Preview

Screenshot from 2021-03-01 12-02-45

As part of #957


thatStrangeGuyThorchain (approved):
LGTM

PR #1011, opened 2021-03-01
Add missing pool shares

### LINKED ISSUE #1009: [LTC] Add missing pool shares
https://user-images.githubusercontent.com/61792675/109519807-e1ba2080-7aab-11eb-8c15-6aec05f145b1.mp4

thatStrangeGuyThorchain:
- fixed missed pools shares by creating data streams based on ENABLED_CHAINS const

closes #1009

### DIFF
--- a/src/renderer/services/midgard/shares.ts
+++ b/src/renderer/services/midgard/shares.ts
@@ -8,6 +8,7 @@ import * as Rx from 'rxjs'
 import * as RxOp from 'rxjs/operators'
 
 import { THORCHAIN_DECIMAL } from '../../helpers/assetHelper'
+import { eqPoolShare } from '../../helpers/fp/eq'
 import { liveData, LiveData } from '../../helpers/rx/liveData'
 import { triggerStream } from '../../helpers/stateHelper'
 import { MemberPool } from '../../types/generated/midgard'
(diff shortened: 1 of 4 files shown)


src/renderer/services/midgard/shares.ts
```diff
@@ -121,7 +122,12 @@ const createSharesService = (
 
   // Loads and combines `PoolShare`'s by given addresses
   const loadCombineSharesByAddresses$ = (addresses: Address[]): PoolSharesLD =>
-    FP.pipe(addresses, A.map(combineShares$), liveData.sequenceArray, liveData.map(A.flatten))
+    FP.pipe(
+      addresses,
+      A.map(combineShares$),
+      liveData.sequenceArray,
+      liveData.map(FP.flow(A.flatten, A.uniq(eqPoolShare)))
```
veado, line 129: Why do we need `A.uniq` here ^?

src/renderer/views/wallet/PoolShareView.tsx
```diff
@@ -40,23 +40,27 @@ export const PoolShareView: React.FC = (): JSX.Element => {
...
     () =>
       FP.pipe(
-        sequenceTOption(oBitcoinAddress, oBinanceAddress),
-        O.fold(() => Rx.EMPTY, combineSharesByAddresses$)
+        ENABLED_CHAINS,
+        A.map(addressByChain$),
+        (addresses) => Rx.combineLatest(addresses),
+        RxOp.switchMap(
+          FP.flow(
+            A.filterMap(FP.identity),
```
veado, line 53: Why `A.filterMap(FP.identity)` ^ ?

veado (approved):

src/renderer/views/wallet/PoolShareView.tsx
thatStrangeGuyThorchain: previously we had 
```ts
 sequenceTOption(oBitcoinAddress, oBinanceAddress),
 O.fold(() => Rx.EMPTY, combineSharesByAddresses$)
```
but during development not every chain address is `O.some('stringAddress')` but there might me `O.none` which would fail all sequence of `sequenceTOption(oBitcoinAddress, oBinanceAddress...),` which is unacceptable. with `filterMap(FP.identity)` we filter up `O.none` values and unwrap values to the plain `string[]` at a single place

veado (approved):

Issue #994, opened 2021-02-26
[BCH] Add Send component


veado:
As part of #821

PR #1012, opened 2021-03-01
[BCH] Add business logic for fees

veado:
and update fee handling of LTC + BTC to be in common with BCH

Needed for #994


thatStrangeGuyThorchain (dismissed):
Almost the same changes as I have at my swap task 😅

PR #1018, opened 2021-03-02
[BCH] swap/deposit/withdraw

thatStrangeGuyThorchain:
implements  BCH:
- swap
- withdraw sym
- deposit sym


src/renderer/services/chain/fees/deposit.ts
```diff
@@ -85,7 +89,7 @@ const depositFeeByChain$ = (chain: Chain, type: DepositType): FeeLD => {
     case PolkadotChain:
       return Rx.of(RD.failure(Error('Deposit fee for Polkadot has not been implemented')))
     case BCHChain:
-      return Rx.of(RD.failure(Error('Deposit fee for Bitcoin Cash has not been implemented')))
+      return BCH.fees$().pipe(liveData.map(({ fast }) => fast))
     case LTCChain:
       return LTC.fees$().pipe(liveData.map(({ fast }) => fast))
   }
```
veado, line 95: FYI: `depositFees$` (line 94)  needs to be fixed with another PR - see #1027

src/renderer/services/chain/fees/deposit.ts
```diff
@@ -31,6 +32,7 @@ export const reloadFees = () => {
   BNB.reloadFees()
```
veado, line 32: FYI: `reloadFees` needs to be fixed with another PR - see #1027

veado (approved):
Good stuff :+1: 

Pls check two notes for another PR.

PR #1023, opened 2021-03-02
support ETH to swap

sarawut11:
- [x] Update `Swap` component handle dynamic fees
- [x] Update swap service to handle dynamic fees
- [x] New types `SwapFeeParams`, `SwapFeesParams`, `SwapFeesHandler`, `LoadSwapFeesHandler` etc.
- [x] Add `convertToBase8` helper 
- [x] Use latest @xchainjs/xchain-ethereum@0.13.2

closes #998

PR #1030, opened 2021-03-03
Use latest xchain-eth@0.13.2

veado:


veado:
Close in favour of #1023

PR #1033, opened 2021-03-04
[Deposit] Fix fees to depend on memo etc

### LINKED ISSUE #1027: [Deposit] Fix fees to depend on memo etc.
## Issue

Chain fees might based on amount, memo (BCH/LTC/BTC) or based on amount, memo, receipient(ETH), which needs to be addressed in `epositFees$`, `reloadFeesByChain` etc. in [services/chain/fees/deposit.ts](https://github.com/thorchain/asgardex-electron/blob/342cd39cf5d9e2ff85d3528971d7d84ca2676d6b/src/renderer/services/chain/fees/deposit.ts).

Also, DON'T [reload all fees at once for all chains](https://github.com/thorchain/asgardex-electron/blob/342cd39cf5d9e2ff85d3528971d7d84ca2676d6b/src/renderer/services/chain/fees/deposit.ts#L31-L36) - it needs to be fixed.

## Reference:

- Matrix of [Multichain fees](https://github.com/thorchain/asgardex-electron/issues/640) 
- Notes about [MULTICHAIN TXS + MEMOS + FEES](https://docs.google.com/document/d/1YkZ_51NNmFdGRQZmfwv6yWAO3BNjls8ssO_1L5gfqz8/edit#heading=h.mpcg7i4jfnb7)

- [Fix for `swapFee$`](https://github.com/thorchain/asgardex-electron/pull/1023/commits/8eafc43605b63ea96a7c7152869447207448ecc4#diff-3d7d6a9de441adfaeb62a2929d9272d03939c4ab4f7c44645f2077387f21513fR95-R112)

thatStrangeGuyThorchain:
- fees/deposit.ts
--  `depositFees$` - refactored to receive parameters needed to calculate fees
-- `depositFeeByChain$` - refactored to receive parameters needed to calculate fees; refactored to get memo purely from parameters; added support for `ETH` fees

- ~`AsymDeposit.tsx` and `SymDeposit.tsx` upgraded to re-request in case chain or amount to deposit was changed. to avoid creation of multiple requests [`setDepositFees`](https://github.com/thorchain/asgardex-electron/pull/1033/files#diff-13b694da7eb491527fdeb8d2323767d9d0cbd39fcb4af79e1669b6b5c701cb3bR119) of `useObservableState`. simple usage of `useObservableState(() => dataStream$)` will have no any effect in case if deposit params were changed during component's lifecycle as [`useObservableState` will call the epic-like `init` function only once](https://github.com/crimx/observable-hooks/blob/master/packages/observable-hooks/src/use-observable-state.ts#L33) so for updating data for re-requesting fees [`useEffect` was used](https://github.com/thorchain/asgardex-electron/pull/1033/files#diff-13b694da7eb491527fdeb8d2323767d9d0cbd39fcb4af79e1669b6b5c701cb3bR146-R159)~
- added resetting amount to deposit in case changing asset and after finishing tx
- removed callback for reloading all fees. it was unused so this is safe remove
closes #1027

### DIFF
--- a/src/renderer/components/deposit/add/AsymDeposit.tsx
+++ b/src/renderer/components/deposit/add/AsymDeposit.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useState } from 'react'
+import React, { useCallback, useEffect, useMemo, useState } from 'react'
 
 import * as RD from '@devexperts/remote-data-ts'
 import { PoolData } from '@thorchain/asgardex-util'
@@ -14,7 +14,9 @@ import { Col } from 'antd'
 import BigNumber from 'bignumber.js'
 import * as FP from 'fp-ts/lib/function'
 import * as O from 'fp-ts/lib/Option'
+import { useObservableState } from 'observable-hooks'
 import { useIntl } from 'react-intl'
+import * as RxOp from 'rxjs/operators'
 
 import { Network } from '../../../../shared/api/types'
 import { ZERO_BASE_AMOUNT, ZERO_BN } from '../../../const'
@@ -22,10 +24,17 @@ import { isChainAsset } from '../../../helpers/assetHelper'
 import { sequenceTOption } from '../../../helpers/fpHelpers'
 import { useSubscriptionState } from '../../../hooks/useSubscriptionState'
 import { INITIAL_ASYM_DEPOSIT_STATE } from '../../../services/chain/const'
-import { Memo, DepositFeesRD, AsymDepositState, AsymDepositStateHandler } from '../../../services/chain/types'
+import {
+  Memo,
+  AsymDepositState,
+  AsymDepositStateHandler,
+  LoadDepositFeesHandler,
+  DepositFeesHandler,
+  DepositFeesParams,
+  DepositFeesRD
+} from '../../../services/chain/types'
 import { PoolAddress } from '../../../services/midgard/types'
 import { ValidatePasswordHandler } from '../../../services/wallet/types'
-import { DepositType } from '../../../types/asgardex'
 import { PasswordModal } from '../../modal/password'
 import { TxModal } from '../../modal/tx'
 import { DepositAssets } from '../../modal/tx/extra'
@@ -43,8 +52,8 @@ export type Props = {
   poolAddress: O.Option<PoolAddress>
   memo: O.Option<Memo>
   priceAsset?: Asset
-  fees: DepositFeesRD
-  reloadFees: (type: DepositType) => void
+  reloadFees: LoadDepositFeesHandler
+  fees$: DepositFeesHandler
   reloadBalances: FP.Lazy<void>
   viewAssetTx: (txHash: string) => void
   validatePassword$: ValidatePasswordHandler
@@ -75,7 +84,7 @@ export const AsymDeposit: React.FC<Props> = (props) => {
     priceAsset,
     reloadFees,
     reloadBalances = FP.constVoid,
-    fees,
+    fees$,
     onChangeAsset,
     disabled = false,
     deposit$,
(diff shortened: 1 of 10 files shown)


src/renderer/services/chain/fees/deposit.ts
```diff
@@ -59,65 +62,120 @@ export const reloadFees$: Rx.Observable<O.Option<LoadFeesHandler>> = selectedPoo
   RxOp.map(O.map(reloadFeesByChain))
 )
 
-// State to reload deposit fees
-const { get$: reloadDepositFees$, set: reloadDepositFees } = observableState<DepositType>('asym')
+// Trigger stream to reload deposit fees
+const { stream$: reloadDepositFees$, trigger: reloadDepositFees } = triggerStream()
```
thatStrangeGuyThorchain, line 66: we create new streams in case chain or amount was changed so there is no need to handle `reloadDepositFees` with any parameters

src/renderer/components/deposit/add/AsymDeposit.tsx
```diff
@@ -104,14 +112,34 @@ export const AsymDeposit: React.FC<Props> = (props) => {
...
+    [asset, assetAmountToDeposit, oMemo, oPoolAddress]
+  )
+
+  const [depositFeesRD] = useObservableState<DepositFeesRD>(() => chainFees$(depositFeesParams), RD.initial)
+
+  // reload fees whenever params have been changed
+  useEffect(() => {
+    reloadFees(depositFeesParams)
+  }, [depositFeesParams, reloadFees])
+
```
veado, line 134: Pls check latest changes of `swap$` - we reload fees **after** changing assets or input values (amount/memo) now. With it we avoid some uneeded requests to get fees. _It can be done with another PR..._

src/renderer/services/chain/fees/deposit.ts
```diff
@@ -60,64 +57,136 @@ export const reloadFees$: Rx.Observable<O.Option<LoadFeesHandler>> = selectedPoo
...
+         */
+        return (
+          // Check if entered chain was changed
+          eqChain.equals(prev.asset.chain, next.asset.chain) &&
+          // Check if entered amount was changed
+          eqBaseAmount.equals(prev.amount, next.amount)
+        )
+      }
+      return false
+    }),
```
veado, line 151: How about to provide a `eqDepositFeesParams` helper?

veado (dismissed):
LGTM! Just two suggestions.

src/renderer/components/deposit/add/AsymDeposit.tsx
thatStrangeGuyThorchain: yep. here is the same idea but in another place. at the `fees/deposit.ts` (your next comment) with `distinctUntilChanged`. just because deposit fees  are used in 2 separate places (sym+asym) it was moved to the service-layer to have the same logic for different places

src/renderer/services/chain/fees/deposit.ts
thatStrangeGuyThorchain: yeah. no problem. just as it used in a single place have not done it. will add it in few minutes

src/renderer/components/deposit/add/AsymDeposit.tsx
thatStrangeGuyThorchain: created #1042

veado: `distinctUntilChanged  is good, but not enough. Imagine an user change his amount in the input field. With any small change, a request to load fees will be done. We can minimize such requests by adding a `onBlur` handler to the input to re-load fees only, if the user "leaves" the input.

veado: Resolved by #1044

Issue #1024, opened 2021-03-03
ADD: Swap Page UX


thorchain-admin:
# Output Amount

1) Output Amount font should use the same font as input amount
2) Remove `(LTC)` from the output, it is implied already
3) Make `(LTC)` as simply `LTC` on the right-side swap rate estimations, and the fee label

<img width="753" alt="image" src="https://user-images.githubusercontent.com/41812729/109750891-31742580-7c2d-11eb-9e1f-205564027ccc.png">

1) Remove (

# Improve Token Row

1) Make the Row have both CHAIN and TICKER (look at Asgard.Exchange)
2) Make the font Larger
3) Add more horizontal margin between the logo and CHAIN/TICKER

<img width="363" alt="image" src="https://user-images.githubusercontent.com/41812729/109750944-54063e80-7c2d-11eb-8633-bb1f4ef335bd.png">

PR #1034, opened 2021-03-04
Improve Token Row

junkai121:
- [x] Output Amount font should use the same font as input amount
- [x] Remove (LTC) from the output, it is implied already
- [x] Make the Row have both CHAIN and TICKER (look at Asgard.Exchange)
- [x] Make the font Larger
- [x] Add more horizontal margin between the logo and CHAIN/TICKER

All part of #1024


veado (approved):
Nice!

PR #1035, opened 2021-03-04
[Swap] Improve handling of fees, max amount, reloding fees etc.

veado:
- [x] Reload fees: 
    - After changing slider
    - After changing input value
    - After switching assets
- [x] Store previous source/target - needed to reload fees while changing   
- [x] New helper: `sourceAssetAmount`, `sourceChainAsset`, `sourceChainAssetAmount`, `maxAmountToSwap`
- [x] "Reset" values after switching target / source assets
- [x] Add `pendingSwitchAssets` state to avoid race conditions of balances + fees while switching assets (see #1037)
- [x] Provide default `maxAmountToSwap` for not logged in users to play around with swap
- [x] Add `onBlurHandler` to `AssetInput`

All other changes just movements of existing code in `Swap`

### DIFF
--- a/src/renderer/components/swap/Swap.tsx
+++ b/src/renderer/components/swap/Swap.tsx
@@ -13,7 +13,9 @@ import {
   baseAmount,
   formatAssetAmount,
   formatAssetAmountCurrency,
-  delay
+  delay,
+  assetAmount,
+  assetToBase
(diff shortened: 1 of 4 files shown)


junkai121 (approved):
LGTM!

PR #1039, opened 2021-03-04
Prepare v0.0.7

veado:
+ enable BCH
+ Override minimizer of webpack (TerserPlugin) to fix sending BCH txs in production build


craco.config.js
```diff
@@ -1,4 +1,6 @@
+// const { whenProd } = require('@craco/craco')
```
thatStrangeGuyThorchain, line 1: remove this line pls

thatStrangeGuyThorchain (approved):
Nice!

craco.config.js
veado: Good catch. I will remove it with another PR to avoid a re-build of releases just for this small thing.

PR #1046, opened 2021-03-05
fix swap for ERC20 tokens show and ETH target fee

sarawut11:


src/renderer/helpers/walletHelper.ts
```diff
@@ -32,7 +32,9 @@ export const getWalletBalanceByAsset = (
     O.chain(([walletBalances, asset]) =>
       FP.pipe(
         walletBalances,
-        A.findFirst(({ asset: assetInList }) => eqAsset.equals(assetInList, asset))
+        A.findFirst(({ asset: assetInList }) =>
+          eqAsset.equals({ ...assetInList, symbol: assetInList.symbol.toUpperCase() }, asset)
```
veado, line 36: It seems `eqAsset` needs to be updated to following:

```ts
export const eqAsset: Eq.Eq<Asset> = {
  equals: (x, y) =>
    eqString.equals(x.chain, y.chain) && eqString.equals(x.symbol.toUpperCase(), y.symbol.toUpperCase())
}
``` 

Then you can revert changes to:
```ts
eqAsset.equals(assetInList, asset))
```

veado (changes requested):
LGTM! One suggestion.

veado (approved):

Issue #1015, opened 2021-03-02
ADD: UX improvements


thorchain-admin:
# ~Refresh button on add Liquidity~ done #1050 

when adding liquidity, need refresh button top right to let user refresh to see liquidity update

<img width="1297" alt="image" src="https://user-images.githubusercontent.com/41812729/109614030-ee10ad00-7b81-11eb-9e61-7df4ea19fa8f.png">

# ~SELF tag for owned addresses~ done #1080

For any addresses that are owned by user - replace the address with `SELF` instead of address

<img width="1355" alt="image" src="https://user-images.githubusercontent.com/41812729/109614095-11d3f300-7b82-11eb-908d-0a99b09980a4.png">

# ~CHAIN tag for pools~ done #1062 

Show pools as `CHAIN.ASSET` instead of just `ASSET`, for ETH/BNB assets, strip out the `-ID` at the end

<img width="1306" alt="image" src="https://user-images.githubusercontent.com/41812729/109614309-624b5080-7b82-11eb-845e-d32b64fa9a10.png">

# ~NIGHT MODE ICON~ done #1058

use:
<img width="280" alt="image" src="https://user-images.githubusercontent.com/41812729/109618245-39798a00-7b87-11eb-8967-640eee020af2.png">

https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020?node-id=3521%3A25751

# ~Pool Filtering~ done #1079 

Pill buttons below heading, 

* `BASE` -> all the base assets, eg BTC.BTC, ETH.ETH etc
* `BEP2` -> all BNB assets
* `ERC20` -> all ETH tokens
* clear button -> if filter on, then show this button, which removes filter

<img width="638" alt="image" src="https://user-images.githubusercontent.com/41812729/109619335-96297480-7b88-11eb-886d-e4fbf851c4b3.png">

thatStrangeGuyThorchain:
all issues are resolved. see appropriate PRs

PR #1080, opened 2021-03-09
SELF tag for owned addresses

### LINKED ISSUE #1052: SELF tag for owned addresses
as a part of #1015

thatStrangeGuyThorchain:
- `AddressEllipsis` - added `displayText` prop
- added self tag for owned address

closes #1052
as a part of #1015


src/renderer/components/wallet/txs/table/TxsTable.tsx
```diff
@@ -51,12 +59,25 @@ export const TxsTable: React.FC<Props> = (props): JSX.Element => {
...
-      <Styled.Text key={key}>
-        <AddressEllipsis address={address} chain={chain} network={network} />
-      </Styled.Text>
-    ),
-    [chain, network]
+    (address: Address, key: string) => {
+      const displayText = FP.pipe(
+        oWalletAddres,
+        O.map((walletAddress) =>
+          walletAddress === address ? (
```
veado, line 66: Just a double check: Are both addresses from same kind ("ui addresses" vs. "addresses")? Just to make sure we are comparing same kind of addresses. 

Pls check if it still works for ETH.

src/renderer/components/uielements/addressEllipsis/AddressEllipsis.tsx
```diff
@@ -19,10 +19,11 @@ type Props = {
   className?: string
   enableCopy?: boolean
   linkIcon?: React.ReactElement
+  displayText?: string | React.ReactElement
```
veado, line 22: What's the reason to extend `AddressEllipsis` instead of using a lable for "SELF" addresses? Or do we need a copy icon for SELF addresses ...?

thatStrangeGuyThorchain: hm. makes sense

src/renderer/components/wallet/txs/table/TxsTable.tsx
thatStrangeGuyThorchain: yep. we compare the same addresses
image
image
image

veado (approved):

PR #1094, opened 2021-03-12
Improve data handling for `PoolAssetDetail`s

veado:
To avoid re-parsing asset strings provided by Midgard in different places again and again, this PR fixes it to do it only once in `services/midgard/pools`. Change is needed to simplify Swap state handling (more with another PR).

Changes:

- [x] Introduce `PoolAssetDetail` to get rid of `AssetWithPrice` 
- [x] Rename `AssetDetail` -> `PoolAssetDetail` etc.
- [x] Introduce `getPoolAssetDetail` + `getPoolAssetsDetail` (incl. test)

### DIFF
--- a/src/renderer/services/midgard/types.ts
+++ b/src/renderer/services/midgard/types.ts
@@ -27,19 +27,13 @@ export type PoolAssets = Asset[]
 export type PoolAssetsRD = RD.RemoteData<Error, PoolAssets>
 export type PoolAssetsLD = LiveData<Error, PoolAssets>
 
-export type AssetDetail = {
-  asset: string
-  dateCreated: number
-  priceRune: string
+export type PoolAssetDetail = {
+  asset: Asset
+  assetPrice: BigNumber
 }
 
-export type AssetDetails = AssetDetail[]
-export type AssetDetailsRD = RD.RemoteData<Error, AssetDetails>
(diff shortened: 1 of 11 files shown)


src/renderer/services/midgard/types.ts
```diff
@@ -27,19 +27,13 @@ export type PoolAssets = Asset[]
 export type PoolAssetsRD = RD.RemoteData<Error, PoolAssets>
 export type PoolAssetsLD = LiveData<Error, PoolAssets>
 
-export type AssetDetail = {
-  asset: string
-  dateCreated: number
-  priceRune: string
+export type PoolAssetDetail = {
+  asset: Asset
+  assetPrice: BigNumber
```
thatStrangeGuyThorchain, line 32: actually we have [AssetWithAmount](https://github.com/thorchain/asgardex-electron/blob/develop/src/renderer/types/asgardex.ts#L12) type. maybe there is a sense to use it ?

thatStrangeGuyThorchain (approved):

src/renderer/services/midgard/types.ts
veado: Yep, type is almost similar. However, `assetPrice` means price for this asset, but not amount of it. Also we might want to add other properties to `PoolAssetDetail` later (similar to originally `MidgardPoolDetail`).

PR #1102, opened 2021-03-15
Fix Decimal issue on swap page.

### LINKED ISSUE #1101: [ETH] Decimal issue on swap page.
Go to ETH/RUNE or ETH/BNB swap page.
User can input ETH amount which is larger than the account balance.
Even it's able to drag to swap.

It seems a decimal handling issue.

junkai121:
close #1101 

- make sure the target amount decimal is the same as source asset decimal.


src/renderer/components/swap/Swap.tsx
```diff
@@ -332,6 +332,9 @@ export const Swap = ({
 
   const setAmountToSwap = useCallback(
     (targetAmount: BaseAmount) => {
+      // make sure the target amount decimal is the same as source asset decimal.
+      targetAmount.decimal = sourceAssetAmount.decimal
```
veado, line 336: Imagine: 
- source: ETH (18 decimal)
- target: BTC (8 decimal)

With this change ^ BTC will have 18 decimal ...

veado (changes requested):

src/renderer/components/swap/Swap.tsx
junkai121: Here, the targetAmount is the user input for source asset.

veado: Ah, I've missed that ^ ...

veado (approved):
Good find!

PR #1109, opened 2021-03-15
Filter only RUNE for Thorchain

### LINKED ISSUE #1108: Asset view issue for thorchain tokens
In assets view, we have 3 assets for thorchain (RUNE, BNB/BNB, ETH/ETH).
Once I click BNB/BNB or ETH/ETH, the page goes blank.

junkai121:
close #1108 

- filter only AssetRuneNative when getting balances.

### DIFF
--- a/src/renderer/services/thorchain/balances.ts
+++ b/src/renderer/services/thorchain/balances.ts
@@ -1,3 +1,5 @@
+import { AssetRuneNative } from '@xchainjs/xchain-util'
+
 import { triggerStream } from '../../helpers/stateHelper'
 import * as C from '../clients'
 import { client$ } from './common'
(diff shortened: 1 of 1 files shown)


src/renderer/services/thorchain/balances.ts
```diff
@@ -6,6 +8,6 @@ import { client$ } from './common'
 const { stream$: reloadBalances$, trigger: reloadBalances } = triggerStream()
 
 // State of balances loaded by Client
-const balances$: C.WalletBalancesLD = C.balances$(client$, reloadBalances$)
+const balances$: C.WalletBalancesLD = C.balances$(client$, reloadBalances$, [AssetRuneNative])
```
veado, line 11: Add a comment (something like this):
```
Currently `AssetRuneNative` is supported only. Remove asset list if we want to get balances of all assets at THORChain.
```

junkai121: ok

veado (approved):

PR #1119, opened 2021-03-17
fix ETH txStatus

### LINKED ISSUE #1118: [ETH] deposit tx dialog doesn't reflect actual tx status
image

The tx was actually confirmed but the deposit tx dialog doesn't reflect it correctly.
The reason is that it recognizes ETH tx as an ERC20 one.

sarawut11:
closes #1118


veado (approved):
Good find :+1:

Issue #1136, opened 2021-03-18
AssetLabel component


veado:
We currently use `AssetData` to display 

```
TICKER
CHAIN
```

for example in Pools Overview. 

However, `AssetData` is a component with stuff we don't need (`icon`, `price`, `network` etc.). Instead of using `AssetData`, provide a component to display ticker + chain only. Call it `AssetLabel` and use it wherever we display `ticker` + `chain`.

PR #1137, opened 2021-03-18
Add ticker + chain in withdraw page

### LINKED ISSUE #1126: [WITHDRAW] Add ticker + chain to asset icon
## HAVE

Screenshot from 2021-03-17 16-38-41

## WANT

```
--------
| icon |  TICKER     amount (price)
|      |  CHAIN
--------
```

junkai121:
close #1126


src/renderer/components/deposit/withdraw/Withdraw.tsx
```diff
@@ -325,6 +329,10 @@ export const Withdraw: React.FC<Props> = ({
 
       <Styled.AssetContainer>
         <Styled.AssetIcon asset={asset} network={network} />
+        <Styled.Label nowrap>
+          <Styled.TickerLabel className="ticker">{asset.ticker}</Styled.TickerLabel>
+          <Styled.TickerLabel className="small">{asset.chain}</Styled.TickerLabel>
```
veado, line 334: Pls provide `Styled.TickerLabel` + `Styled.ChainLabel`. Both are very similar styled components, but we should not to override it by using `className` in other components. Imagine we change ".small" to ".chain" or so, all overridden styles will be lost.

veado: Let do it with another PR after @sarawut11 has finished #1136. _Approving now..._

veado (approved):

src/renderer/components/deposit/withdraw/Withdraw.tsx
junkai121: Just fixed and pushed

veado (approved):

src/renderer/components/deposit/withdraw/Withdraw.tsx
veado: _It seems we can use `AssetData` here. Anyway, this change is not needed yet..._

PR #1153, opened 2021-03-19
add liquidity fixes

### LINKED ISSUE #1131: ADD Liquidity Fail First Time
# Add Liquidity First Time

When adding liquidity first time, and after TX is successful:

* reset the slider to 0%
* reset the input fields to 0
* spinner icon in "you don't have any shares", wait 5 seconds, then auto-refresh from Midgard to show pool shares
* make sure the `REFRESH` button works (it currently doesn't)

image

thatStrangeGuyThorchain:
`services/midgard/shares`:
- fixed wrong error handling
- replaced triggerStream with observableState for setting delay on demand

`SymDeposit` and `AsymDeposit`:
  - added resetting form values on form closed
  - fixed zero max amount to deposit while fees are reloading
  - added auto reloading shares after 5 seconds after deposit tx

`WidthdrawDepositView`:
  - added re-requesting shares after withdraw

`Widthdraw`:
  - added resetting withdraw amount at onFinishTxModal
  
  closes #1131


src/renderer/services/midgard/shares.ts
```diff
@@ -21,7 +21,13 @@ const createSharesService = (
...
   const api$ = byzantine$.pipe(RxOp.map(RD.map(getMidgardDefaultApi)))
 
-  const { stream$: reloadShares$, trigger: reloadShares } = triggerStream()
+  /**
+   * We need a possibility to reload shares info with delay for this
+   * purpose we need to use observableState instead of plain triggerStream here
+   * @example after sending deposit tx there is no sense to reload shares info
+   *          right after deposit-tx was succeed
+   */
+  const { get$: reloadSharesWithTimer$, set: reloadSharesWithTimer } = observableState<number | undefined>(0)
```
veado, line 30: `observableState<number>(0)` ?

src/renderer/services/midgard/shares.ts
```diff
@@ -37,14 +43,36 @@ const createSharesService = (
    */
   const shares$ = (address: Address): PoolSharesLD =>
     FP.pipe(
-      Rx.combineLatest([api$, reloadShares$]),
-      RxOp.map(([api]) => api),
+      Rx.combineLatest([api$, reloadSharesWithTimer$]),
+      RxOp.switchMap(([api, delayTime]) =>
```
veado, line 47: `delayTime = 0` in case you wan't to support `observableState<number | undefined>(0)`

veado (changes requested):
Code LGTM! Just two small things ...

src/renderer/services/midgard/shares.ts
thatStrangeGuyThorchain: actually this is a thing that i wanted to support: to have [reloadShares](https://github.com/thorchain/asgardex-electron/pull/1153/files#diff-0ddf52bae472e9088bc56e84ec464a0c5b090d17e60db08e03b35675e0ac8f6bR154) as a function with non-required parameter to use it just as a single function everywhere where no need to have a delay. for example for the manual [reload button](https://github.com/thorchain/asgardex-electron/blob/develop/src/renderer/views/deposit/DepositView.tsx#L103)

`Rx.timer` has a `0` value inside by default 
image

anyway i've moved [default value for the reloadShares](https://github.com/thorchain/asgardex-electron/pull/1153/commits/818fa6c4ced52178617ead69b143d6f85efdce29)

veado: > actually this is a thing that i wanted to support: to have reloadShares as a function with non-required parameter to use it just as a single function everywhere where no need to have a delay.

Make sense! So why have you changed it with last commit to `observableState<number>(0)`?

> Rx.timer has a 0 value inside by default

True. However, by reading your code you can't know if `Rx.timer` excepts `undefined` or not. That's why `Rx.timer(undefined)` looks weird at first view and there is nothing wrong to give extra information using `delayTime = 0` then. Anyway, just a nitpick...

veado (approved):

src/renderer/services/midgard/shares.ts
thatStrangeGuyThorchain: >Make sense! So why have you changed it with 818fa6c to observableState<number>(0)?
just to make things clearer as you suggested and pass a `0` value explicitly with
```    reloadShares: (delayTime = 0) => reloadSharesWithTimer(delayTime),```

Issue #1130, opened 2021-03-18
[EPIC] UX Cleanup


thorchain-admin:
# Night/Day Icon

Currently is on right, too much clutter

<img width="333" alt="image" src="https://user-images.githubusercontent.com/41812729/111562733-3c5ab880-87de-11eb-9b24-2e56347393f7.png">

Move to left, add NIGHT icon like BEPSwap

<img width="345" alt="image" src="https://user-images.githubusercontent.com/41812729/111562800-55636980-87de-11eb-97ff-52b5a85037f4.png">

#  Filters

Use the following filters:

`BASE | USD | BEP2 | ERC20`

BASE: gas assets
USD: All coins with `USD` in ticker
BEP2: BNB + BEP2
ERC20: ETH + ERC20

<img width="505" alt="image" src="https://user-images.githubusercontent.com/41812729/111562829-61e7c200-87de-11eb-9e47-126cd397ede8.png">

# Styles on Setting

Increase font size of `SETTINGS` to be larger than sub-heading

<img width="359" alt="image" src="https://user-images.githubusercontent.com/41812729/111563041-c145d200-87de-11eb-93b9-1ff075b120aa.png">

# Balances when asset is on wallet

When searching asset, show balance if present

<img width="310" alt="image" src="https://user-images.githubusercontent.com/41812729/111563145-eafef900-87de-11eb-8b01-b3f05325e3f8.png">

# Decimal Precision

Do not display any more than 2 decimals, anywhere, unless the amount is less than 2 decimals, in which case only show 2 sigfig. 

<img width="369" alt="image" src="https://user-images.githubusercontent.com/41812729/111574750-9e261d00-87f4-11eb-9098-84159326c53a.png">

PR #1155, opened 2021-03-21
UX Cleanup

sarawut11:
part of #1130 

- Night/Day Icon Update
- Filters Update
- Styles on Setting


veado (approved):
Code LGTM

package.json
```diff
@@ -108,6 +108,7 @@
     "react-circular-progressbar": "^2.0.3",
     "react-dom": "^17.0.1",
     "react-draggable": "^4.4.3",
+    "react-feather": "^2.0.9",
```
veado, line 111: Don't introduce another component for just two icons. You can import svg as we have done it with other icons (see `src/renderer/components/icons/index.ts`

veado (changes requested):

package.json
sarawut11: @veado I couldn't find moon icon in figma and also in antd. The ticket said to use bepswap moon icon so I checked bepswap code and noticed that they used above component rather than svg.
Can we keep it for now or are you okay to use moon svg file from the google?

veado: Pls use a svg file (will be provided via Slack).

veado (approved):

PR #1166, opened 2021-03-23
[Swap] Fix decimal for locked / not imported wallet users

### LINKED ISSUE #1165: [Swap] Get correct decimal for assets
Currently we use decimal based on user balances, but it does not work for users with "unlocked" / "not imported" wallets.

veado:
Main changes:
- [x]  Get decimal using `xchain-eth` helper `getDecimal` + cache result 
- [x] Use Latest xchain-eth@0.16.0

Quick fixes:
- [x] Extract `toThorNetwork`
- [x]  DepositView: Get selectedPoolAsset from stream, not from route
- [x] `selectedPoolAddresses$` -> `selectedPoolAddress$`

Fix #1165

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "@xchainjs/xchain-client": "^0.7.0",
     "@xchainjs/xchain-cosmos": "^0.11.0",
     "@xchainjs/xchain-crypto": "^0.2.3",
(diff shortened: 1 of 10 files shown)


thatStrangeGuyThorchain (commented):
LGTM

PR #1192, opened 2021-03-25
support asset busd-74e for pricing

sarawut11:


veado (changes requested):
Don't remove `AssetBUSDBAF`, just add another `AssetBUSD74E` to that list

veado (approved):
LGTM

PR #1206, opened 2021-03-27
Introduce `PoolOverviewDetail` type

veado:
## TL;DR
Rename `PoolDetail` to `PoolOverviewDetail`

## Summary
To get data for pool overview we currently do requests to:
- `pools` to get `PoolDetails` for all pools
-  `v2/history/waps/` to get more data for each pool

We do need this way for pool overview only, but not for other places where we need data of `PoolDetail` as well. To keep the differences, we do need another type `PoolOverviewDetail` besides Midgards `PoolDetail`.


sarawut11 (dismissed):
Code looks good to me!

Issue #1216, opened 2021-03-29
Upgrade dependencies


veado:
```bash

ncu 

Checking asgardex-electron/package.json
[====================] 109/109 100%

 @ant-design/icons                       ^4.3.0  →     ^4.6.2     
 @craco/craco                            ^6.0.0  →     ^6.1.1     
 @devexperts/remote-data-ts              ^2.0.4  →     ^2.0.5     
 @ledgerhq/hw-app-btc                   ^5.38.0  →    ^5.46.0     
 @ledgerhq/hw-transport-node-hid        ^5.38.0  →    ^5.46.0     
 @openapitools/openapi-generator-cli    ^2.1.16  →     ^2.2.4     
 @polkadot/api                           ^3.9.3  →     ^4.2.1     
 @polkadot/util                          ^5.6.3  →     ^6.0.5     
 @xchainjs/xchain-bitcoincash            ^0.7.1  →     ^0.7.2     
 antd                                   ^4.10.2  →    ^4.14.1     
 axios                                  ^0.21.0  →    ^0.21.1     
 bitcore-lib-cash                       ^8.24.2  →    ^8.25.7     
 bitcore-mnemonic                       ^8.24.2  →    ^8.25.7     
 cosmos-client                          ^0.39.2  →    ^0.42.2     
 electron-log                            ^4.3.1  →     ^4.3.2     
 ethers                                  ^5.0.9  →    ^5.0.32     
 fp-ts                                   ^2.9.3  →     ^2.9.5     
 fp-ts-rxjs                             ^0.6.11  →    ^0.6.13     
 fs-extra                                ^9.0.1  →     ^9.1.0     
 observable-hooks                        ^3.1.2  →     ^4.0.3     
 react                                  ^17.0.1  →    ^17.0.2     
 react-circular-progressbar              ^2.0.3  →     ^2.0.4     
 react-dom                              ^17.0.1  →    ^17.0.2     
 react-intl                            ^5.10.15  →    ^5.15.3     
 react-scripts                           ^4.0.1  →     ^4.0.3     
 react-test-renderer                    ^17.0.1  →    ^17.0.2     
 rxjs                                    ^6.6.3  →     ^6.6.7     
 typescript                              ^4.1.3  →     ^4.2.3     
 @react-theming/storybook-addon          ^1.0.3  →     ^1.1.1     
 @storybook/addon-actions               ^6.1.12  →    ^6.1.21     
[...]

PR #1228, opened 2021-03-29
Remove dependencies for Polkadot (temporary)

veado:
It will be added again as soon as we support Polkadot

As part of #1216


junkai121 (dismissed):
LGTM.

PR #1231, opened 2021-03-29
Update dependencies (29-03-21)

veado:
Some "highlights":

- [x] Latest `typescript@4.2.x`
- [x] Add config files for latest `husky`
- [x] Update config for `eslint` 
- [x] Fix `fp-ts` dependencies in `yarn.lock`
- [x] Remove `bitcore-* deps`

Issue #1110, opened 2021-03-15
Re-visit shapeshift/hdwallet


veado:
Few months ago it failed to use [shapeshift/hdwallet](https://github.com/shapeshift/hdwallet) in ASGDX (see [Electron Support #229](https://github.com/shapeshift/hdwallet/issues/229)). Shapeshift said it's an upstream issue. Let's give another try by using latest [@ledgerhq/hw-transport-webusb](https://www.npmjs.com/package/@ledgerhq/hw-transport-webusb)

PR #1236, opened 2021-03-30
Upgrade to Electron@12 + Node@14

veado:
- [x] Electron@12
- [x] Node@14
- [x] Upgrade to latest `node-abi` to fix Electron build
- [x] Remove `@ledgerhq-*` (it might be re-added later by re-visiting Ledger stuff - see #1110) 
- [x] Bump to `v0.0.9@alpha.1` (for testing CI only) 

Close #1231


.nvmrc
```diff
@@ -1 +1 @@
-12.18.3
+14.16.0
```
veado, line 1: That's the Node version Electron@12 is using (see https://www.electronjs.org/releases/stable#12.0.0)

thatStrangeGuyThorchain (approved):

Issue #1161, opened 2021-03-23
[EPIC] PoolDetail 


veado:
Similar to BEPSwap (_Current[ Figma layout for ASGDX](https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020?node-id=3678%3A0) will be updated soon..._)

1. Add route 
1. Add skeleton of `PoolDetailView` + `PoolDetail` component
1. Service: Get data of pool details
1. Service: Get data for tx history
1. Service: Get data for chart
1. PoolDetail: Add title + buttons linked to deposit / swap
1. PoolDetail: Show details within tiles on the top
1. PoolDetail: Add tx history (table) 
1. PoolDetail: Add chart

PR #1249, opened 2021-04-01
Pool detail history

thatStrangeGuyThorchain:
`ActivePools`: enabled onCLick for a pool-row

`midgard/pools`: removed filtering out `O.none` values from `selectedPoolDetail$` as it leaded to the wrong caching results

`PoolHistoryView`: added business logic for history loading;

`PoolDetailsView`:
  - added `BackLink` and `RefreshButton`
  - added caching prev results;

`midgard/poolActionsHistory`: added `isHistoryLoading$` data;

as a part of #1161

### DIFF
--- a/src/renderer/components/pool/PoolDetails.tsx
+++ b/src/renderer/components/pool/PoolDetails.tsx
@@ -60,28 +60,30 @@ export const PoolDetails: React.FC<Props> = ({
   )
   return (
     <Styled.Container>
-      <A.Col span={24}>
-        <PoolTitle asset={asset} priceUSD={priceUSD} isLoading={isLoading} />
-      </A.Col>
-      <A.Col xs={24} md={8}>
-        <PoolCards
-          depth={depth}
-          volume24hr={volume24hr}
-          allTimeVolume={allTimeVolume}
-          totalStakers={totalStakers}
-          totalSwaps={totalSwaps}
-          priceSymbol={priceSymbol}
-          depthTrend={depthTrend}
-          volume24hrTrend={volume24hrTrend}
-          allTimeVolumeTrend={allTimeVolumeTrend}
-          totalSwapsTrend={totalSwapsTrend}
-          totalStakersTrend={totalStakersTrend}
-          isLoading={isLoading}
-        />
-      </A.Col>
-      <A.Col xs={24} md={16}>
-        <PoolChart isLoading={isLoading} />
-      </A.Col>
+      <Styled.TopContainer>
+        <A.Col span={24}>
+          <PoolTitle asset={asset} priceUSD={priceUSD} isLoading={isLoading} />
+        </A.Col>
+        <A.Col xs={24} md={8}>
+          <PoolCards
+            depth={depth}
+            volume24hr={volume24hr}
+            allTimeVolume={allTimeVolume}
+            totalStakers={totalStakers}
+            totalSwaps={totalSwaps}
+            priceSymbol={priceSymbol}
+            depthTrend={depthTrend}
+            volume24hrTrend={volume24hrTrend}
+            allTimeVolumeTrend={allTimeVolumeTrend}
+            totalSwapsTrend={totalSwapsTrend}
+            totalStakersTrend={totalStakersTrend}
+            isLoading={isLoading}
+          />
+        </A.Col>
+        <A.Col xs={24} md={16}>
+          <PoolChart isLoading={isLoading} />
+        </A.Col>
+      </Styled.TopContainer>
       <A.Col span={24}>{renderHistoryView}</A.Col>
     </Styled.Container>
   )
--- a/src/renderer/components/pool/PoolDetails.tsx
+++ b/src/renderer/components/pool/PoolDetails.tsx
@@ -8,11 +8,10 @@ import * as O from 'fp-ts/lib/Option'
 import { PoolCards } from './PoolCards'
 import { PoolChart } from './PoolChart'
 import * as Styled from './PoolDetails.style'
-import { PoolHistory } from './PoolHistory'
 import { PoolTitle } from './PoolTitle'
 
 export type Props = {
-  asset: O.Option<Asset>
+  asset: Asset
   depth: AssetAmount
   depthTrend?: BigNumber
   volume24hr: AssetAmount
@@ -27,6 +26,7 @@ export type Props = {
   priceUSD: AssetAmount
   priceSymbol?: string
   isLoading?: boolean
+  HistoryView: React.ComponentType<{ poolAsset: Asset }>
 }
 
 export const PoolDetails: React.FC<Props> = ({
@@ -43,12 +43,13 @@ export const PoolDetails: React.FC<Props> = ({
   allTimeVolumeTrend,
   totalSwapsTrend,
   totalStakersTrend,
-  isLoading
+  isLoading,
+  HistoryView
 }) => {
   return (
     <Styled.Container>
       <A.Col span={24}>
-        <PoolTitle asset={asset} priceUSD={priceUSD} isLoading={isLoading} />
+        <PoolTitle asset={O.some(asset)} priceUSD={priceUSD} isLoading={isLoading} />
       </A.Col>
       <A.Col xs={24} md={8}>
         <PoolCards
@@ -70,7 +71,7 @@ export const PoolDetails: React.FC<Props> = ({
         <PoolChart isLoading={isLoading} />
       </A.Col>
       <A.Col span={24}>
-        <PoolHistory isLoading={isLoading} />
+        <HistoryView poolAsset={asset} />
       </A.Col>
     </Styled.Container>
   )
--- a/src/renderer/components/pool/PoolDetails.tsx
+++ b/src/renderer/components/pool/PoolDetails.tsx
@@ -1,9 +1,10 @@
-import React from 'react'
+import React, { useMemo } from 'react'
 
 import { Asset, AssetAmount } from '@xchainjs/xchain-util'
 import * as A from 'antd'
 import BigNumber from 'bignumber.js'
-import * as O from 'fp-ts/lib/Option'
+import * as FP from 'fp-ts/function'
+import * as O from 'fp-ts/Option'
 
 import { PoolCards } from './PoolCards'
 import { PoolChart } from './PoolChart'
@@ -11,7 +12,7 @@ import * as Styled from './PoolDetails.style'
 import { PoolTitle } from './PoolTitle'
 
 export type Props = {
-  asset: Asset
+  asset: O.Option<Asset>
   depth: AssetAmount
   depthTrend?: BigNumber
   volume24hr: AssetAmount
@@ -46,10 +47,21 @@ export const PoolDetails: React.FC<Props> = ({
   isLoading,
   HistoryView
 }) => {
+  const renderHistoryView = useMemo(
+    () =>
+      FP.pipe(
+        asset,
+        O.fold(
+          () => <></>,
+          (asset) => <HistoryView poolAsset={asset} />
+        )
+      ),
+    [asset, HistoryView]
+  )
   return (
     <Styled.Container>
       <A.Col span={24}>
-        <PoolTitle asset={O.some(asset)} priceUSD={priceUSD} isLoading={isLoading} />
+        <PoolTitle asset={asset} priceUSD={priceUSD} isLoading={isLoading} />
       </A.Col>
       <A.Col xs={24} md={8}>
         <PoolCards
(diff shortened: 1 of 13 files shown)


src/renderer/views/pool/PoolDetailsView.tsx
```diff
@@ -72,38 +77,61 @@ export const PoolDetailsView: React.FC = () => {
...
+  const onRefreshData = useCallback(() => {
+    reloadSelectedPoolDetail()
+    reloadActionsHistory()
+  }, [reloadSelectedPoolDetail, reloadActionsHistory])
+
+  const refreshButtonDisabled = useMemo(() => {
+    return isHistoryLoading || FP.pipe(poolDetailRD, RD.isPending)
+  }, [isHistoryLoading, poolDetailRD])
+
+  const prev = useRef<React.ReactElement>()
```
veado, line 102: Not sure if this is really a good idea to store such a heavy component into a `Ref` ... What's the reason for this? Maybe another way could be storing all props of the component, but not the component itself. Also, `PoolDetailsView` has already a `loading` prop, why not just using it?

src/renderer/views/pool/PoolHistoryView.tsx
```diff
@@ -0,0 +1,110 @@
...
+export const PoolHistory: React.FC<Props> = ({ className, poolAsset }) => {
+  const {
+    service: { poolActionsHistory }
+  } = useMidgardContext()
+
+  const { getExplorerTxUrl$ } = useThorchainContext()
+
+  const prevActionsPage = useRef<O.Option<PoolActionsHistoryPage>>(O.none)
+
+  const requestParams = useRef(DEFAULT_REQUEST_PARAMS)
```
veado, line 37: Why is a reference to `requestParams` needed?

src/renderer/views/pools/ActivePools.tsx
```diff
@@ -176,18 +176,18 @@ export const ActivePools: React.FC = (): JSX.Element => {
             loading={loading}
             rowKey="key"
             // TODO(@asgdx-team): Uncomment when pool detail is ready
```
veado, line 178: ^ can be removed

src/renderer/components/pool/PoolDetails.tsx
```diff
@@ -43,35 +44,47 @@ export const PoolDetails: React.FC<Props> = ({
...
   totalSwapsTrend,
   totalStakersTrend,
-  isLoading
+  isLoading,
+  HistoryView
 }) => {
+  const renderHistoryView = useMemo(
+    () =>
+      FP.pipe(
+        asset,
```
veado, line 53: BTW: Why is `asset` an Option in props of `PoolDetail` and not just an `Asset`. If we don't know the asset, we don't have any knowledge about any other data of a pool, which makes `PoolDetail` useless. And in `PoolDetailView` we render `PoolDetails` only if we have an `asset` ...

src/renderer/views/pool/PoolDetailsView.tsx
```diff
@@ -72,38 +77,61 @@ export const PoolDetailsView: React.FC = () => {
...
+          (e: Error) => {
+            return <PoolStatus label={intl.formatMessage({ id: 'common.error' })} displayValue={e.message} />
+          },
+          ([asset, poolDetail]) => {
+            return (prev.current = (
+              <PoolDetails
+                asset={O.some(asset)}
+                depth={getDepth(poolDetail, priceRatio)}
+                volume24hr={get24hrVolume(poolDetail, priceRatio)}
+                allTimeVolume={getAllTimeVolume(poolDetail, priceRatio)}
```
veado, line 124: Rethink about properties of `PoolDetail`. Many of them accepts `PoolDetail`, why not just passing `PoolDetail` down to `PoolDetails` and do all the calculation there. It will simplify its props and data handling `PoolDetailsView` ...

src/renderer/views/pool/PoolHistoryView.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  return (
+    <PoolActionsHistory
+      className={className}
+      currentPage={requestParams.current.page + 1}
+      actionsPageRD={historyPage}
+      prevActionsPage={prevActionsPage.current}
+      goToTx={goToTx}
+      changePaginationHandler={setCurrentPage}
+      clickTxLinkHandler={goToTx}
```
veado, line 105: Why two different props (`gotToTx` and `clickTxLinkHandler` for same value (`goToTx`)?

veado (changes requested):

src/renderer/services/midgard/poolActionsHistory.ts
```diff
@@ -71,11 +80,19 @@ export const createPoolActionsHistoryService = (
...
         )
-      )
+      ),
+      RxOp.tap((value) => {
+        if (RD.isPending(value)) {
+          setIsHistoryLoading(true)
+        } else {
+          setIsHistoryLoading(false)
+        }
+      })
```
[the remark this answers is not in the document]
thatStrangeGuyThorchain: yeah. good question. `PoolHistoryView` is a separate View-component to avoid additional 100+ lines of independent business logic at the `PoolDetailsView` and `PoolDetailsView` actually can not know about actions loading state 'cuz [actions$](https://github.com/thorchain/asgardex-electron/blob/develop/src/renderer/services/midgard/poolActionsHistory.ts#L37) create a stream based on request parameters. but `PoolDetailsView` needs to know if data is loading to disable `Refresh` button...

src/renderer/views/pool/PoolHistoryView.tsx
thatStrangeGuyThorchain: as you [can see below](https://github.com/thorchain/asgardex-electron/pull/1249/files#diff-3fc93a20db589613c143bc252a64976566d4c6511c0ace3395293b2fc14a0eccR46) we save previously set params inside of `useObservableState`'s callback to have an opportunity to have small callbacks like [setCurrentPage](https://github.com/thorchain/asgardex-electron/pull/1249/files#diff-3fc93a20db589613c143bc252a64976566d4c6511c0ace3395293b2fc14a0eccR62) which provide **only changed** parameters for actions-request

src/renderer/views/pool/PoolDetailsView.tsx
thatStrangeGuyThorchain: yes. it makes sense. but here this values was not changed at all (just re-formatted) just because there are another PRs with changes for this area. i have touched that just to avoid conflicts with sibling PRs and would like to refactor this props-interface with another PR.

src/renderer/views/pool/PoolHistoryView.tsx
thatStrangeGuyThorchain: mistake from history-page PR. redundant of course...

src/renderer/services/midgard/poolActionsHistory.ts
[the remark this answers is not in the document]
veado: Why not just push `actions$` or a `isLoading` state down it to `PoolDetailsView`?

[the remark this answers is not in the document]
veado: Another way is to provide `const actionsIsLoading$ = actions$.pipe(RxOp.map(RD.isPending))` or something.

src/renderer/components/pool/PoolDetails.tsx
thatStrangeGuyThorchain: yes. the same thought what came into my head here and i've tried to change it but something went wrong (dont remember what exactly) ([change commit](https://github.com/thorchain/asgardex-electron/pull/1249/commits/f274373c1428907dcde08cfaac50e56f3f0f3e15), and [revert one](https://github.com/thorchain/asgardex-electron/pull/1249/commits/7367c24822c82a33d42a89f5a582a36496b4640e)). anyway now i've tried one more time. works fine

src/renderer/views/pool/PoolHistoryView.tsx
veado: Why don't you use `DEFAULT_REQUEST_PARAMS` all the time to update it with `asset` and / or `page`?

src/renderer/views/pool/PoolDetailsView.tsx
```diff
@@ -107,28 +101,39 @@ export const PoolDetailsView: React.FC = () => {
...
-
-            return <PoolDetails {...prevProps.current} />
-          }
+        oRouteAsset,
+        O.fold(
+          () => (
+            <PoolStatus
+              label={intl.formatMessage({ id: 'common.error' })}
+              displayValue={intl.formatMessage({ id: 'routes.invalid.asset' }, { asset })}
+            />
```
veado, line 110: You can use `ErrorView` here (like in other views for same purpose) and `PoolStatus` might be not needed anymore (and can be removed?).

src/renderer/services/midgard/poolActionsHistory.ts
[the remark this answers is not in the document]
thatStrangeGuyThorchain: > Why not just push actions$ or a isLoading state down it to PoolHistoryView?

if we do this we have to handle all request parameters at the `PoolDetailsView` directly to create an `actions$` stream (as i've mentioned before [actions$](https://github.com/thorchain/asgardex-electron/blob/develop/src/renderer/services/midgard/poolActionsHistory.ts#L37) if a function which depends on parameters)

>Another way is to provide const actionsIsLoading$ = actions$.pipe(RxOp.map(RD.isPending)) or something.

we can not do this for the same reason as `actions$` is a function. not a separate stream. seems naming is a little bit wrong...

i can create `actions$` as an observable state at the [poolActionsHistory](https://github.com/thorchain/asgardex-electron/blob/develop/src/renderer/services/midgard/poolActionsHistory.ts) but in this case we have to be careful with possible reduntant values left at this stream and reset stream's values manually when any component uses this "hot" stream.
if you say do things ^this^ way - i will do

src/renderer/views/pool/PoolHistoryView.tsx
thatStrangeGuyThorchain: mmm. 'cuz they are default 😅 for example:
- user set a filter
- user wants to navigate to another page
- in case using `DEFAULT_REQUEST_PARAMS` filter will be restored to default

src/renderer/views/pool/PoolDetailsView.tsx
thatStrangeGuyThorchain: yes. changed to use `ErrorView` but can not remove it as PoolStatus used at another place too
image

src/renderer/views/pool/PoolHistoryView.tsx
veado: Ok, I see. Please comment to the code to understand the logic behind the handling of `Partial<LoadActionsParams>` & Co. (which is nice, but not clear at a first view...)

src/renderer/services/midgard/poolActionsHistory.ts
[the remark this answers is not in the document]
veado: > if we do this we have to handle all request parameters at the PoolDetailsView directly to create an actions$ stream (as i've mentioned before actions$ if a function which depends on parameters)

It will be updated/fixed by going this way ^, but with another PR (as discussed in DM)

src/renderer/views/pool/PoolDetailsView.tsx
```diff
@@ -72,38 +75,62 @@ export const PoolDetailsView: React.FC = () => {
...
+                  return <ErrorView title={message} />
+                },
+                (poolDetail) => {
+                  prevProps.current = {
+                    depth: getDepth(poolDetail, priceRatio),
+                    volume24hr: get24hrVolume(poolDetail, priceRatio),
+                    allTimeVolume: getAllTimeVolume(poolDetail, priceRatio),
+                    totalSwaps: 0 /* getTotalSwaps(history) */,
+                    totalStakers: 0 /* getTotalStakers(poolDetail) */,
+                    priceUSD: getPriceUSD(poolDetail, priceRatio),
```
veado, line 123: Similar to [previous message](https://github.com/thorchain/asgardex-electron/pull/1249#discussion_r605742418): Just push `poolDetail` into `PoolDetail` and do the rest to get `depth`, `volume24hr`, `allTimeVolume` etc, there.

thatStrangeGuyThorchain: yep. will be changed with another PR to avoid conflicts resolving (as discussed in DMs)

src/renderer/views/pool/PoolDetailsView.tsx
veado: Ok

veado (approved):

PR #1252, opened 2021-04-01
add business logic for depth history

sarawut11:


src/renderer/services/midgard/pools.ts
```diff
@@ -689,6 +692,47 @@ const createPoolsService = (
...
+      RxOp.shareReplay(1)
+    )
+  }
+
+  const { stream$: reloadDepthHistory$, trigger: reloadDepthHistory } = triggerStream()
+
+  const getDepthHistory$ = (params: GetDepthHistoryParams): DepthHistoryLD =>
+    FP.pipe(
+      Rx.combineLatest([selectedPoolAsset$, reloadDepthHistory$]),
+      RxOp.filter(([oSelectedPoolAsset, _]) => O.isSome(oSelectedPoolAsset)),
```
veado, line 717: Pls remove this line ^ . _I know we have it similar at other places, but @thatStrangeGuyThorchain has realized today, it breaks data flow in a stream and we already handle the case of `O.none` in line 723.

veado (changes requested):

veado (approved):

PR #1255, opened 2021-04-02
add pool detail chart

sarawut11:


src/renderer/components/pool/PoolChart.tsx
```diff
@@ -1,17 +1,169 @@
...
+import { LiquidityData, VolumeData } from './types'
 
 export type Props = {
+  selectedPricePoolAsset: SelectedPricePoolAsset
   isLoading?: boolean
+  priceRatio: BigNumber
+  volumeAllTimeData: O.Option<VolumeData[]>
+  volumeWeekData: O.Option<VolumeData[]>
+  liquidityAllTimeData: O.Option<LiquidityData[]>
+  liquidityWeekData: O.Option<LiquidityData[]>
```
thatStrangeGuyThorchain, line 22: actually as i can see this values are the same at all...
types
```ts
export type VolumeData = {
  time: number
  poolVolume: string
}

export type LiquidityData = {
  time: number
  runeDepth: string
}
```
can be converted to utilized interface like
```ts
type ChartItem = {
  time: number; // which is actually timeStamp...
  value: string;
}
```

the second thing is why do we need all of this separation of props ?
the main question here is what will we do in case we want to display more data at the chart ? pass over and over new pairs (for week and all time periods) again and again ? and what if there are more time periods to select ? day/week/month/3 months/ 6 months ? really dont like this...
from my point of view `PoolChart`(and `DefaultChart`) should receive only values the need at the moment like:
```ts
type ChartProps = {
  isLoading: boolean
  displayType: 'line' | 'bar'
  charData: ChartItem[]
  timePeriods: string[] // or ('all' | 'week')[]
  setTimePeriod: (period: string) => void
  dataTypes: string[] // Liquidity and Volume
  setDataType: (dataType: string) => void
}
```
with ^^^this^^^ we can reuse such chart component anywhere we wanted to and decide which exact data we want to display on demand

so with such changes (if they are approved) we just need to place calculation and storing of all this data to the View-layer.
actually it also might be utilized. for `PoolChartView.tsx` might be something like this (this is not a final code. just main idea):

```ts
const availableTimePeriods = ['week', 'all'] as const
type TimePeriod = typeof availableTimePeriods[number]
const availableDataTypes = ['liquidity',  'volume'] as const
type DataType = typeof availableDataTypes[number]
type DataRequestParams = { period: TimePeriod; dataType: DataType }
  const savedParams = useRef<DataRequestParams>({
    period: availableTimePeriods[0],
    dataType: availableDataTypes[0]
  })
  const [chartDataRD, updateChartData] = useObservableState<RD.RemoteData<any, any>, Partial<A>>(
    (params$) =>
      FP.pipe(
        params$,
        RxOp.map((params) => (savedParams.current = { ...savedParams.current, ...params })),
        RxOp.switchMap((params) => {
          const getDataStream$ = params.dataType === 'liquidity' ? getDepthHistory$ : getSwapHistory$
// ^^^  getDepthHistory$ and getSwapHistory$  getters should be modified to return a single data interfaces ^^^
          const requestParams = params.period === 'all' ? {} : { from: weekAgoTime, to: curTime }

          return getDataStream$({
            ...requestParams,
            interval: 'day' as const
          })
        }),
        RxOp.map(/* map data as you need to */)
      ),
    RD.initial
  )

  const setTimePeriod = useCallback(
    (period: TimePeriod) => {
      updateChartData({ period })
    },
    [updateChartData]
  )

  const setDataTypePeriod = useCallback(
    (dataType: DataType) => {
      updateChartData({ dataType, period: availableTimePeriods[0] })
    },
    [updateChartData]
  )
```

so in this case for adding any new `DataType` or `TimePeriod` we just need to configure appropriate constants and modify appropriate getters... and Chart component will take ONLY data it needs to display....

@veado  what do you think ?

thatStrangeGuyThorchain (changes requested):

src/renderer/views/pool/PoolChartView.helper.ts
```diff
@@ -0,0 +1,24 @@
+import moment from 'moment'
+
+export const getEoDTime = () => {
```
veado, line 3: Could you add a comment what `EoD` is?

src/renderer/components/pool/PoolDetails.tsx
```diff
@@ -62,7 +65,7 @@ export const PoolDetails: React.FC<Props> = ({
           />
         </A.Col>
         <A.Col xs={24} md={16}>
-          <PoolChart isLoading={isLoading} />
+          <ChartView isLoading={isLoading} priceRatio={priceRatio} />
```
veado, line 68: Why do we need `isLoading` here? If I'm not wrong, data for `ChartView` depends on `getSwapHistory$` and `getDepthHistory`, which are handled by `ChartView` itself. These both streams are including a `pending` state.

src/renderer/views/pool/PoolChartView.tsx
```diff
@@ -0,0 +1,119 @@
...
+      pools: { selectedPricePoolAsset$, getSwapHistory$, getDepthHistory$ }
+    }
+  } = useMidgardContext()
+
+  const selectedPricePoolAsset = useObservableState<SelectedPricePoolAsset>(selectedPricePoolAsset$, O.none)
+
+  const curTime = getEoDTime()
+  const weekAgoTime = getWeekAgoTime()
+
+  const [swapAllHistoryRD] = useObservableState(
```
veado, line 32: Currently data for `swapAllHistoryRD` (all data) and `swapWeekHistoryRD` (weekly data) is loaded with two request. Since an user can switch between `all` and `weekly` data,  but maybe he will never switch), my suggestion is to load data only we really show in the chart (weekly OR all data). But that's more an optimization thing we don't need atm.

src/renderer/views/pool/PoolChartView.tsx
```diff
@@ -0,0 +1,119 @@
...
+    RD.toOption,
+    O.map((history) =>
+      history.intervals.map((interval) => ({
+        time: Number(interval.startTime),
+        poolVolume: interval.totalVolume
+      }))
+    )
+  )
+
+  const [depthAllHistoryRD] = useObservableState(
```
veado, line 70: Same here: Currently we load `depthAllHistoryRD` and `depthWeekHistoryRD` together, but need just one of them (depending on users selection: `all` or `weekly` data). That's an optimization we can add later.

veado (changes requested):
Two questions (1, 2) and some thoughts (1, 2)

veado:
Closed in favour of  #1261

src/renderer/views/pool/PoolChartView.tsx
veado: Fixed by #1261

src/renderer/views/pool/PoolChartView.tsx
veado: Fixed by #1261

src/renderer/components/pool/PoolChart.tsx
veado: Fixed by #1261

PR #1259, opened 2021-04-03
Support ETH/ERC20 for MCCN

veado:
- [x] Use latest `xchain-eth@0.17.0`
- [x] Update dependencies (`ethers`, `xchain-utils`, `xchain-ltc`)
- [x] Extend balance handling for ETH/ERC20 to support MCCN
- [x] Fix: ETH/ERC20 balances of previous selected network are still visible 
- [x] Add ENVs needed for Ethplorer


junkai121 (approved):
LGTM!

src/renderer/services/wallet/balances.ts
```diff
@@ -148,15 +148,43 @@ const btcLedgerBalance$ = FP.pipe(
...
+/**
+ * Setter to return cached stream for ETH balances (mainnet)
+ * to use one (and same) stream created by `ETH.balances$` factory only
+ */
+const getEthBalancesMainnet$ = () => {
+  if (!ethBalancesMainnet$) {
+    ethBalancesMainnet$ = ETH.balances$().pipe(RxOp.shareReplay(1))
+  }
+  return ethBalancesMainnet$
+}
```
thatStrangeGuyThorchain, line 177: hm... it makes sense from a perspective to show to user some data ASAP. but dont we want to re-request balances when user switches network ? from my point of view it makes more sense....

veado: Currently we have two different streams (or better "factories") for ETH balances for `testnet` and `mainnet` `getEthBalancesTestnet$` vs. `getEthBalancesMainnet$` Whenever a user switches to another network, a factory is called. The problem is, if an user switch back and forth (testnet <-> mainnet), many streams will be created if we don't cache it in that way. And you are right, in this case we show cached balances which might not be up-to-date.

As discussed in DM, the team will try to find another approach, but with another PR.

PR #1268, opened 2021-04-05
Improve handling of ETH balances by switching network

### LINKED ISSUE #1262: Improve handling of ETH balances by switching network
By switching networks (`testnet` <-> `chaosnet`) streams for ETH balances are hard to manage (see comments [here](https://github.com/thorchain/asgardex-electron/pull/1259#discussion_r607032055) and [here](https://github.com/thorchain/asgardex-electron/pull/1259/files#diff-5739933d3907c82ee03cf50350bd964fddfe07b1e797decb23beca177fa29eb4R8-R11)). 

Whether we will find another solution for handling streams of ETH balances or we just reload the application by switching the network.

thatStrangeGuyThorchain:
- `services/clients/balances.ts:balances$`: removed `shareReplay`-based caching strategy
- `services/wallet/balances.ts`: added memory-based cache strategy. main logic is:
- - cached value for every chain will be updated separately for every succeed data loading.
- - when there is no cached data for  a chain data will be requested from api
- - if there is “force” reload (by clicking refresh button or  calling reloadBalanceByChain at the swap/deposit views) data for appropriate chain will be re-requested and appropriate cache will be updated (so every next reading from a stream will always have latest loaded data)
- - if network was changed cached values are dropped for every chain (manual `network$.subscribe`)

closes #1262

### DIFF
--- a/src/renderer/contexts/WalletContext.tsx
+++ b/src/renderer/contexts/WalletContext.tsx
@@ -15,7 +15,7 @@ import {
   getTxs$,
   setSelectedAsset,
   resetTxsPage,
-  clearSavedBalances
+  resetWalletBalancesState
 } from '../services/wallet'
 
 type WalletContextValue = {
@@ -30,7 +30,7 @@ type WalletContextValue = {
   getTxs$: typeof getTxs$
   setSelectedAsset: typeof setSelectedAsset
   resetTxsPage: typeof resetTxsPage
-  clearSavedBalances: typeof clearSavedBalances
+  resetWalletBalancesState: typeof resetWalletBalancesState
 }
 
 const initialContext: WalletContextValue = {
@@ -45,7 +45,7 @@ const initialContext: WalletContextValue = {
   getTxs$,
   setSelectedAsset,
   resetTxsPage,
-  clearSavedBalances
+  resetWalletBalancesState: resetWalletBalancesState
 }
 const WalletContext = createContext<Option<WalletContextValue>>(none)
 
--- a/src/renderer/contexts/WalletContext.tsx
+++ b/src/renderer/contexts/WalletContext.tsx
@@ -14,7 +14,8 @@ import {
   getExplorerTxUrl$,
   getTxs$,
   setSelectedAsset,
-  resetTxsPage
+  resetTxsPage,
+  clearSavedBalances
 } from '../services/wallet'
 
 type WalletContextValue = {
@@ -29,6 +30,7 @@ type WalletContextValue = {
   getTxs$: typeof getTxs$
   setSelectedAsset: typeof setSelectedAsset
   resetTxsPage: typeof resetTxsPage
+  clearSavedBalances: typeof clearSavedBalances
 }
 
 const initialContext: WalletContextValue = {
@@ -42,7 +44,8 @@ const initialContext: WalletContextValue = {
   selectedAsset$,
   getTxs$,
   setSelectedAsset,
-  resetTxsPage
+  resetTxsPage,
+  clearSavedBalances
 }
 const WalletContext = createContext<Option<WalletContextValue>>(none)
 
--- a/src/renderer/contexts/WalletContext.tsx
+++ b/src/renderer/contexts/WalletContext.tsx
@@ -14,8 +14,7 @@ import {
   getExplorerTxUrl$,
   getTxs$,
   setSelectedAsset,
-  resetTxsPage,
-  resetWalletBalancesState
+  resetTxsPage
 } from '../services/wallet'
 
 type WalletContextValue = {
@@ -30,7 +29,6 @@ type WalletContextValue = {
   getTxs$: typeof getTxs$
   setSelectedAsset: typeof setSelectedAsset
   resetTxsPage: typeof resetTxsPage
-  resetWalletBalancesState: typeof resetWalletBalancesState
 }
 
 const initialContext: WalletContextValue = {
@@ -44,8 +42,7 @@ const initialContext: WalletContextValue = {
   selectedAsset$,
   getTxs$,
   setSelectedAsset,
-  resetTxsPage,
-  resetWalletBalancesState: resetWalletBalancesState
+  resetTxsPage
 }
 const WalletContext = createContext<Option<WalletContextValue>>(none)
 
--- a/src/renderer/services/binance/balances.ts
+++ b/src/renderer/services/binance/balances.ts
@@ -1,14 +1,10 @@
-import { observableState } from '../../helpers/stateHelper'
+import { triggerStream } from '../../helpers/stateHelper'
 import * as C from '../clients'
 import { client$ } from './common'
 
-const { get$: reloadBalances$, set: setReload } = observableState<'trigger' | ''>('')
-
-const reloadBalances = (state: 'trigger' | '' = 'trigger') => {
-  setReload(state)
-}
+const { stream$: reloadBalances$, trigger: reloadBalances, reset: resetReload } = triggerStream()
 
 // State of balances loaded by Client
 const balances$: C.WalletBalancesLD = C.balances$(client$, reloadBalances$)
 
-export { balances$, reloadBalances, reloadBalances$ }
+export { balances$, reloadBalances, reloadBalances$, resetReload }
--- a/src/renderer/services/binance/balances.ts
+++ b/src/renderer/services/binance/balances.ts
@@ -1,11 +1,17 @@
-import { triggerStream } from '../../helpers/stateHelper'
+import { observableState } from '../../helpers/stateHelper'
 import * as C from '../clients'
 import { client$ } from './common'
 
+const { get$: reloadBalances$, set: setReload } = observableState<'trigger' | ''>('')
+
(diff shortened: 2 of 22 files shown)


src/renderer/views/wallet/SettingsView.tsx
```diff
@@ -337,13 +337,21 @@ export const SettingsView: React.FC = (): JSX.Element => {
     getLedgerErrorDescription(btcLedgerAddress, 'BTC')
   }, [bnbLedgerAddress, btcLedgerAddress, intl, removeLedgerAddress])
 
+  const setNetwork = useCallback(
+    (network: Network) => {
+      clearSavedBalances()
```
veado, line 343: Let's handle network on service layer. Whenever an user change the network, `network$` will trigger a change and we can reset balance state at service layer.

src/renderer/services/wallet/balances.ts
```diff
@@ -34,29 +35,107 @@ export const reloadBalances: FP.Lazy<void> = () => {
...
+      return (state) => THOR.reloadBalances(state)
     case 'LTC':
-      return LTC.reloadBalances
+      return (state) => LTC.reloadBalances(state)
     default:
       return () => {}
   }
 }
 
+let balanceRecord: Partial<Record<Chain, WalletBalancesRD>> = {}
```
veado, line 66: `balanceRecord` -> `walletBalancesState` ?

src/renderer/services/thorchain/balances.ts
```diff
@@ -1,14 +1,20 @@
 import { AssetRuneNative } from '@xchainjs/xchain-util'
 
-import { triggerStream } from '../../helpers/stateHelper'
+import { observableState /*, triggerStream*/ } from '../../helpers/stateHelper'
 import * as C from '../clients'
 import { client$ } from './common'
 
+const { get$: reloadBalances$, set: setReload } = observableState<'trigger' | ''>('')
```
veado: Maybe just a `boolean` for reload state? `observableState<boolean>(false)`

src/renderer/services/wallet/balances.ts
```diff
@@ -34,29 +35,107 @@ export const reloadBalances: FP.Lazy<void> = () => {
...
+      return LTC.reloadBalances$
+    default:
+      return Rx.EMPTY
+  }
+}
+
+const getChainBalance$ = (chain: Chain): WalletBalancesLD => {
+  const reload$ = FP.pipe(
+    getChainBalanceReload$(chain),
+    RxOp.finalize(() => {
```
veado, line 93: [question] In which case can we expect this stream will be completed?

src/renderer/services/wallet/balances.ts
```diff
@@ -34,29 +35,86 @@ export const reloadBalances: FP.Lazy<void> = () => {
...
+  return getServiceByChain(chain)?.reloadBalances || (() => {})
+}
+
+const resetReloadByChain: (chain: Chain) => void = (chain) => {
+  return getServiceByChain(chain)?.resetReload()
+}
+
+let balanceRecord: Partial<Record<Chain, WalletBalancesRD>> = {}
+
+export const clearSavedBalances = () => {
```
veado, line 72: `clearSavedBalances` -> `resetWalletBalancesState`

src/renderer/services/wallet/balances.ts
```diff
@@ -34,29 +35,86 @@ export const reloadBalances: FP.Lazy<void> = () => {
...
+}
+
+const getChainBalanceData$: (chain: Chain) => WalletBalancesLD = (chain) => {
+  if (chain === 'ETH') {
+    return FP.pipe(
+      network$,
+      RxOp.switchMap((network) => ETH.balances$(network === 'testnet' ? ETHAssets : undefined))
+    )
   }
+  return (getServiceByChain(chain)?.balances$ || Rx.of(RD.initial)) as WalletBalancesLD
```
veado, line 83: `Rx.of<WalletBalancesLD>(...)`

src/renderer/services/wallet/balances.ts
```diff
@@ -34,29 +35,86 @@ export const reloadBalances: FP.Lazy<void> = () => {
...
+      resetReloadByChain(chain)
+    })
+  )
+
+  return FP.pipe(
+    reload$,
+    RxOp.switchMap((reloadTrigger) => {
+      const savedResult = balanceRecord[chain]
+      if (reloadTrigger === '' && savedResult) {
+        if (savedResult) {
```
veado, line 106: Why another guard to check `savedResult` ?

src/renderer/services/wallet/balances.ts
```diff
@@ -34,29 +35,86 @@ export const reloadBalances: FP.Lazy<void> = () => {
   BCH.reloadBalances()
 }
 
-export const reloadBalancesByChain: (chain: Chain) => FP.Lazy<void> = (chain) => {
+const getServiceByChain = (chain: Chain) => {
```
veado, line 38: I would not get all service here, just functions we do really need (such as `reloadBalances`, `reloadBalances$`, `resetReload`)

src/renderer/services/thorchain/balances.ts
thatStrangeGuyThorchain: i've added a `reset` callback to the `triggerStream` helper with [this commit](https://github.com/thorchain/asgardex-electron/pull/1268/commits/74cb0a3383e53facf8a0d508598a703edd4bea8e). what do you think ?

src/renderer/services/wallet/balances.ts
thatStrangeGuyThorchain: as i can see it (and adding some `console.log` expression inside of `finalize` confirms it) as we have `triggerStream` as a `BehaviourSubject` it will be completed on unsubscribe and in all of our cases unsubscribe will be initiated on any View unmount (comment for that is already added)

src/renderer/services/wallet/balances.ts
thatStrangeGuyThorchain: yeah. i had implemented it in such way before but there was so much duplication (for reload/reload$/reset there were the same switch/cases) and i've decided to extract it to the separate function. actually... as i can see when we need to implement a new chain we will pass only a new chain-service to the `getServiceByChain` and everything should work as expected.
anyway ^that^ just thoughts. if you wish i will change that to separate switch/cases for every  of reload/reload$/reset getters

src/renderer/services/wallet/balances.ts
thatStrangeGuyThorchain: yep. just missed

src/renderer/views/wallet/SettingsView.tsx
veado: > we will have circular dependency
If we have a circular dependency, let's re-organize modules

src/renderer/contexts/WalletContext.tsx
```diff
@@ -45,7 +45,7 @@ const initialContext: WalletContextValue = {
   getTxs$,
   setSelectedAsset,
   resetTxsPage,
-  clearSavedBalances
+  resetWalletBalancesState: resetWalletBalancesState
```
veado, line 48: [nitpick] `resetWalletBalancesState: resetWalletBalancesState` -> `resetWalletBalancesState`

src/renderer/views/wallet/SettingsView.tsx
veado: > he other option is to set "infinite" subscription at the wallet/balances like

> ```
> export const clearSavedBalances = () => {
>   balanceRecord = {}
> }
> network$.subscribe(clearSavedBalances)
> ```

This looks good ^ btw.

src/renderer/services/binance/balances.ts
```diff
@@ -1,11 +1,24 @@
...
 import { client$ } from './common'
 
-// `TriggerStream` to reload `Balances`
-const { stream$: reloadBalances$, trigger: reloadBalances } = triggerStream()
+/**
+ * `ObservableState` to reload `Balances`
+ * Sometimes we need to have a way to understand if it simple "load" or "reload" action
+ * e.g. @see src/renderer/services/wallet/balances.ts:getChainBalance$
+ */
+const { get$: reloadBalances$, set: setReloadBalances } = observableState<ReloadType>('load')
```
thatStrangeGuyThorchain, line 11: (here and further) we can not use `observableState<boolean>` as ` C.balances$` requires `reloadStream$` to be `TriggerStream` type which is actually `Observable<string>`

src/renderer/services/wallet/balances.ts
```diff
@@ -34,29 +36,126 @@ export const reloadBalances: FP.Lazy<void> = () => {
   BCH.reloadBalances()
 }
 
-export const reloadBalancesByChain: (chain: Chain) => FP.Lazy<void> = (chain) => {
+type ChainService = {
+  reloadBalances: FP.Lazy<void>
+  resetReload: FP.Lazy<void>
```
veado, line 41: [nitpick] `resetReload` -> `resetReloadBalances`

veado (approved):
Awesome - fantastic work :+1:

PR #1283, opened 2021-04-08
add refresh button in swap page

### LINKED ISSUE #1245: [Swap] Add reload button
to re-load balances + pool data (similar to deposit page)

chaindbkey:
close #1245

### DIFF
--- a/src/renderer/views/swap/SwapView.tsx
+++ b/src/renderer/views/swap/SwapView.tsx
@@ -13,8 +13,7 @@ import * as Rx from 'rxjs'
 import { Network } from '../../../shared/api/types'
 import { ErrorView } from '../../components/shared/error/'
 import { Swap } from '../../components/swap'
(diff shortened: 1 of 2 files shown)


src/renderer/views/swap/SwapView.tsx
```diff
@@ -177,6 +176,12 @@ export const SwapView: React.FC<Props> = (_): JSX.Element => {
     reloadBalances()
   }, [reloadBalances])
 
+  const reloadHandler = useCallback(() => {
+    reloadBalances()
```
veado, line 180: Reload balances of selected assets to swap only

veado (changes requested):

src/renderer/views/swap/SwapView.tsx
veado: We already have it implemented, I've missed it.

veado (approved):
LGTM

PR #1290, opened 2021-04-08
[DepositView] reduce redundant pool-details request

### LINKED ISSUE #1286: [DepositView] reduce redundant pool-details request
for some reason there are duplicated request for pool details at the deposit view. need to try to avoid them
image

thatStrangeGuyThorchain:
Issue description:
`DepositView` is a composition view which render `ShareView` and `Sym/Asym-Deposit` / `Sym/Asym-Withdraw` views. Both of rendered views (`ShareView` + any of listed) are subscribing `selectedPoolDetail$` ([ShareView](https://github.com/thorchain/asgardex-electron/blob/develop/src/renderer/views/deposit/share/ShareView.tsx#L37),  [WithdrawView](https://github.com/thorchain/asgardex-electron/blob/develop/src/renderer/views/deposit/withdraw/WithdrawDepositView.tsx#L46) and etc.). so this leads to duplicated requests.

Solution:
- moved loading of `poolDetails` to the top-level at the `DepositView`

closes #1286

### DIFF
--- a/src/renderer/components/deposit/Deposit.stories.tsx
+++ b/src/renderer/components/deposit/Deposit.stories.tsx
@@ -15,6 +15,7 @@ import { Default as Withdraw } from './withdraw/Withdraw.stories'
 
 const defaultProps: DepositProps = {
   asset: { asset: AssetBNB, decimal: BNB_DECIMAL },
+  poolDetail: RD.initial,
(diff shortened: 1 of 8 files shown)


veado (approved):
Nice find + fix :+1:

PR #1291, opened 2021-04-08
Add approve fee handling for swap page

junkai121:


src/renderer/services/chain/fees/approve.ts
```diff
@@ -0,0 +1,27 @@
...
+        Rx.from(
+          // asset
+          ETH.approveTxFee$(approveParams || params)
+        )
+      )
+    })
+  )
+}
+
+export { reloadApproveFee, approveFee$ }
```
veado, line 27: No need to have `approveFee$` in `chain/fees/approve.ts`, since it's needed for ETH only. It can be removed, but `{ get$: reloadApproveFee$, set: reloadApproveFee }` needs to be move into `src/renderer/services/ethereum/fees.ts` before

veado (approved):
:+1:

Issue #1301, opened 2021-04-09
ADD: FundsCap


thorchain-admin:
The network has `MAXLIQUIDITYRUNE` which is set on mimir

This is the maximum RUNE the network will accept in the pools.

## Counter
1) Display a counter which tracks `totalPooledRune / MAXLIQUIDITYRUNE`

totalPooledRune
https://midgard.thorchain.info/v2/network

MAXLIQUIDITYRUNE
https://thornode.thorchain.info/thorchain/mimir

Put this counter somewhere at the top, above the pools Table. 

<img width="370" alt="image" src="https://user-images.githubusercontent.com/41812729/114250232-63249d00-99e0-11eb-8b7e-a3f7e3b90bdb.png">

## Stop Add Liquidity

1) When the network hits 90% of the `MAXLIQUIDITYRUNE` then disable the ADD LIQUIDITY screen (hide it, or block the button).

PR #1311, opened 2021-04-12
Business logic to get funds cap

veado:
- [x] Get data from `mimir` endpoint
- [x] Introduce `useFundsCapHook` for sharing data between different views (aka ViewModel)
- [x] Prepare `Header` to get data (UI will be part of another PR)

Part of #1301


thatStrangeGuyThorchain (approved):
ilke

PR #1315, opened 2021-04-13
[Swap] cache (store) previously successfully loaded fees

### LINKED ISSUE #1314: [Swap] cache (store) previously successfully loaded fees
right now when any swap amount is changed we have re-loading of fees. fees' pending state leads to the incorrect re-calculating of `max amount to swap`. 

https://user-images.githubusercontent.com/67793194/114514219-b77b8700-9c43-11eb-8f06-b4ff872fe752.mov

thatStrangeGuyThorchain:
- Swap.tsx: added caching of previously successfully loaded fees
closes #1314

### DIFF
--- a/src/renderer/components/swap/Swap.tsx
+++ b/src/renderer/components/swap/Swap.tsx
@@ -39,7 +39,7 @@ import {
 import { getChainAsset, isEthChain } from '../../helpers/chainHelper'
 import { eqAsset, eqBaseAmount, eqOAsset } from '../../helpers/fp/eq'
 import { sequenceSOption, sequenceTOption } from '../../helpers/fpHelpers'
-import { LiveData } from '../../helpers/rx/liveData'
+import { liveData, LiveData } from '../../helpers/rx/liveData'
 import { filterWalletBalancesByAssets, getWalletBalanceByAsset } from '../../helpers/walletHelper'
 import { useSubscriptionState } from '../../hooks/useSubscriptionState'
 import { swap } from '../../routes/pools'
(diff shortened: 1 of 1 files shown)


src/renderer/components/swap/Swap.tsx
```diff
@@ -280,9 +280,18 @@ export const Swap = ({
   const chainFees$ = useMemo(() => fees$, [fees$])
   const approveFees$ = useMemo(() => approveFee$, [approveFee$])
 
+  const chainFeesORef = useRef<O.Option<SwapFees>>(O.none)
```
veado, line 283: [nitpick] `chainFeesORef` -> `prevChainFees` (just to have similar naming as we have at other places)

src/renderer/components/swap/Swap.tsx
```diff
@@ -340,7 +349,8 @@ export const Swap = ({
     async (asset: Asset) => {
       // update `pendingSwitchAssets` state
       setPendingSwitchAssets(true)
-
+      // Reset previously stored fees
+      chainFeesORef.current = O.none
```
veado, line 353: Not in `reset()` ^?

thatStrangeGuyThorchain: i believe no need to do it `reset` method. this `ref`'s value stores last successfully loaded fees for source asset and i suppose we need to reset it only in case `sourceAsset` was changed. `reset` method is called also when target asset is changed and there is no need to reset cached fee value i suppose

veado (approved):
Cool!

PR #1331, opened 2021-04-14
[Wallet] History

### LINKED ISSUE #1321: [Wallet] History
Upgrade `BNB.RUNE` -> `RUNE.RUNE` is displayed as `UNKNOWN` and not linked. (Tested with ASGDX 0.1.0 / chaosnet)

Screenshot from 2021-04-13 18-07-02

thatStrangeGuyThorchain:
- added `SWITCH` TxType;
- `PoolActionsHistory` (and related):
- -  added `availableFilters` prop (`upgrade` filter does not make sense for `PoolDetails` page)
- - fixed non-unique keys for history action (moved to separate `getRowKey` helper + testes)
- - fixed filtering empty `txId`
- `TxType`: added Upgrade icon

closes #1321

### DIFF
--- a/src/renderer/components/poolActionsHistory/PoolActionsHistory.helper.test.ts
+++ b/src/renderer/components/poolActionsHistory/PoolActionsHistory.helper.test.ts
@@ -173,7 +173,8 @@ describe('PoolActionsHistory.helper', () => {
       ).toEqual('inId')
     })
 
-    it('should return default value in case there is no txId (`dateTimrStamp-action.type`)', () => {
+    it('should return default value in case there is no txId (`action.date-action.type`)', () => {
+      // Date(0) is a default date property value for defaultPoolAction
       expect(getRowKey(defaultPoolAction)).toEqual(`${new Date(0)}-SWAP`)
     })
   })
--- a/src/renderer/components/poolActionsHistory/PoolActionsHistory.helper.test.ts
+++ b/src/renderer/components/poolActionsHistory/PoolActionsHistory.helper.test.ts
@@ -3,7 +3,7 @@ import * as O from 'fp-ts/Option'
 
 import { eqAssetsWithAmount } from '../../helpers/fp/eq'
 import { PoolAction, Tx } from '../../services/midgard/types'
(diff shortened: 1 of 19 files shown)


src/renderer/components/poolActionsHistory/PoolActionsHistory.helper.test.ts
```diff
@@ -128,4 +128,53 @@ describe('PoolActionsHistory.helper', () => {
...
+              ...defaultTx,
+              txID: 'outId'
+            }
+          ]
+        })
+      ).toEqual('inId')
+    })
+
+    it('should return default value in case there is no txId (`dateTimrStamp-action.type`)', () => {
+      expect(getRowKey(defaultPoolAction)).toEqual(`${new Date(0)}-SWAP`)
```
veado, line 178: [question] Why `new Date(0)` (which is always the same) and not just `swap` or anything else?

veado: [nitpick] Ah, it's because you use `action.date.toString()}-${action.type}` later. So just add a short note to understand why to use `new Date(0)`  ;)

src/renderer/i18n/de/common.ts
```diff
@@ -67,6 +67,7 @@ const common: CommonMessages = {
   'common.tx.type.refund': 'Erstatten',
   'common.tx.type.deposit': 'Einzahlen',
   'common.tx.type.withdraw': 'Auszahlen',
+  'common.tx.type.upgrade': 'Upgrade - DE',
```
veado: `'common.tx.type.upgrade': 'Upgrade'`,

veado (dismissed):
Nice one!

src/renderer/components/poolActionsHistory/PoolActionsHistory.helper.test.ts
thatStrangeGuyThorchain: hm. it's described at the test-case name. no ? 🤔

veado (approved):

PR #1388, opened 2021-04-29
Support Auto Update

### LINKED ISSUE #889: Support Auto Update
### TODO
1. Simple solution: Check if a new version is available (via release pages) and show an info to users they can download it there
2. Support auto update based on `electron-builder`'s [Auto Update](https://www.electron.build/auto-update).

### Links:
- Blog post:  [Creating and deploying an auto-updating Electron app for Mac and Windows using electron-builder](https://medium.com/@johndyer24/creating-and-deploying-an-auto-updating-electron-app-for-mac-and-windows-using-electron-builder-6a3982c0cee6) 
- Electron Documentaton: [Updating Applications](https://www.electronjs.org/docs/tutorial/updates) 
- [update.electronjs.org](https://github.com/electron/update.electronjs.org) - A free service that makes it easy for open-source Electron apps to update themselves.
- [update-electron-app ](https://github.com/electron/) - A drop-in module that adds autoUpdating capabilities to Electron apps
- Description how [Electron Forge](https://www.electronforge.io/) is dealing with [Auto Update](https://www.electronforge.io/advanced/auto-update)

thatStrangeGuyThorchain:
to test locally play with version at the `package.json` (e.g. "version": "0.0.9")

NOTE:  update-info requests for every application start

closes #889


src/renderer/services/appUpdate/service.ts
```diff
@@ -0,0 +1,20 @@
...
+
+FP.pipe(
+  Rx.from(window.apiAppUpdate),
+  RxOp.catchError(() => Rx.EMPTY),
+  RxOp.map((version) => O.some(version))
+).subscribe(setAppUpdater)
+
+const resetAppUpdater = () => setAppUpdater(O.none)
+
+export { appUpdater$, resetAppUpdater }
```
veado: Put all of this logic of `appUpdate/service.ts ` into a hook `useAppUpdate`

src/renderer/services/appUpdate/service.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { observableState } from '../../helpers/stateHelper'
+
+type AppUpdater = O.Option<string>
+
+const { get$: appUpdater$, set: setAppUpdater } = observableState<AppUpdater>(O.none)
+
+FP.pipe(
+  Rx.from(window.apiAppUpdate),
+  RxOp.catchError(() => Rx.EMPTY),
```
veado: Handle error (we do need to know if something went wrong)

src/renderer/i18n/de/update.ts
```diff
@@ -0,0 +1,9 @@
+import { UpdateMessages } from '../types'
+
+const update: UpdateMessages = {
+  'update.available': 'New version is available - DE',
+  'update.description':
+    'There is a new version {version} available. You can go to the releases page and download it. - DE'
+}
```
veado:
```
'update.available': 'Neues Update verfügbar',
'update.description':
    'Eine neue Version {version} ist verfügbar. Bei Klick auf "Ok" gelangst Du zur Release Webseite, wo Du die neue Version downloaden kannst.'
```

src/renderer/i18n/en/common.ts
```diff
@@ -78,7 +78,8 @@ const common: CommonMessages = {
   'common.time.week': 'week',
   'common.time.all': 'all',
   'common.theme.light': 'Day mode',
-  'common.theme.dark': 'Night mode'
+  'common.theme.dark': 'Night mode',
+  'common.go': 'Go'
```
veado: ```'common.ok' ='Ok'```

src/renderer/i18n/de/common.ts
```diff
@@ -78,7 +78,8 @@ const common: CommonMessages = {
   'common.time.week': 'Woche',
   'common.time.all': 'Gesamt',
   'common.theme.light': 'Tag modus',
-  'common.theme.dark': 'Nacht modus'
+  'common.theme.dark': 'Nacht modus',
+  'common.go': 'Go'
```
veado: 'common.ok' ='Ok'

src/main/preload.ts
```diff
@@ -27,3 +29,7 @@ contextBridge.exposeInMainWorld('apiHDWallet', apiHDWallet)
 // api for storage objects
 contextBridge.exposeInMainWorld('apiCommonStorage', getFileStoreApi('common'))
 contextBridge.exposeInMainWorld('apiUserNodesStorage', getFileStoreApi('userNodes'))
+
+contextBridge.exposeInMainWorld('apiAppUpdate', getApiAppUpdate())
+
+ipcRenderer.on(IPCMessages.LOG, (...args) => console.log('--- log ---', ...args))
```
veado: Still needed: `console.log...` ?

src/renderer/i18n/en/update.ts
```diff
@@ -0,0 +1,8 @@
+import { UpdateMessages } from '../types'
+
+const update: UpdateMessages = {
+  'update.available': 'New version is available',
+  'update.description': 'There is a new version {version} available. You can go to the releases page and download it'
```
veado:
```
'update.available': 'New update available',
'update.description':
    'There is a new version {version} available. Click "Ok" to go to the release page, where you can download it.'
```

veado (changes requested):

src/renderer/services/appUpdate/service.ts
thatStrangeGuyThorchain: ok. in which way ? show something to the user or just internal `console.log` ?

src/main/preload.ts
thatStrangeGuyThorchain: nope

src/renderer/services/appUpdate/service.ts
veado: Show same modal "New version is available", but with an error message (in red) in it's body: "`Check for update failed: ${error.?message ?? error.toString()}`" or similar

src/main/electron.ts
```diff
@@ -140,15 +141,44 @@ const initIPC = () => {
...
+  await autoUpdater.checkForUpdates()
 }
 
+autoUpdater.on('update-available', (info: { version: string }) => {
+  sendUpdateMessage(IPCMessages.UPDATE_AVAILABLE, info.version)
+})
+
+autoUpdater.on('update-not-available', () => {
+  sendUpdateMessage(IPCMessages.UPDATE_NOT_AVAILABLE)
+})
```
veado: Move it into `main/api/appUpdate.ts `

src/main/electron.ts
```diff
@@ -140,15 +141,44 @@ const initIPC = () => {
...
+}
+const sendUpdateMessage: SendUpdateMessage = (message: IPCMessages, version?: string) => {
+  if (mainWindow) {
+    mainWindow.webContents.send(message, version)
+  }
+}
+
+// Disable autoDownload
+autoUpdater.autoDownload = false
+autoUpdater.allowPrerelease = true
```
veado: Move it into `main/api/appUpdate.ts `

src/main/electron.ts
```diff
@@ -140,15 +141,44 @@ const initIPC = () => {
...
+
 const init = async () => {
   await app.whenReady()
   await initMainWindow()
   app.on('window-all-closed', allClosedHandler)
   app.on('activate', activateHandler)
-
   initIPC()
+  await autoUpdater.checkForUpdates()
```
veado: Extract `autoUpdater.checkForUpdates()` into `main/api/appUpdate.ts` to manage requests for checking updates from `renderer` side something like this:
```
// renderer -> main
CHECK_APP_UPDATE

// renderer -> main
APP_UPDATE
```

With this, we can start requests every 6 (or 12hrs etc.) to let users know about important updates, even if an user has open its wallet for two days or so...

package.json
```diff
@@ -1,14 +1,14 @@
...
+  "version": "0.0.1",
   "description": "WALLET AND EXCHANGE CLIENT FOR THORCHAIN",
-  "main": "public/electron.js",
+  "main": "index.js",
   "scripts": {
     "start:web": "craco start",
     "dev": "concurrently --kill-others --success first \"yarn:dev:*\"",
     "dev:app": "cross-env BROWSER=none yarn start:web",
-    "dev:electron": "wait-on http://localhost:3000 && nodemon -I --watch ./src/main/ --exec electron ./public/dev.js",
+    "dev:electron": "wait-on http://localhost:3000 && nodemon -I --watch ./src/main/ --exec electron .",
```
veado, line 11: `--exec electron .` -> `--exec electron ./index.js` (just to make it more clear for non JS users)

thatStrangeGuyThorchain: unfortunately we can not do that for the [same issue](https://github.com/electron/electron/issues/7085#issuecomment-354305046) we discussed with you.
P.S. yep. i've tried to use `./index.js` directly and there is comparison with `electro`'s version
image

src/renderer/views/app/AppView.tsx
```diff
@@ -57,7 +61,10 @@ export const AppView: React.FC = (): JSX.Element => {
   return (
     <Styled.AppWrapper>
       <Styled.AppLayout>
+        {isDesktopView && <AppUpdateView />}
         <Header />
+        {!isDesktopView && <AppUpdateView />}
```
veado:
```ts
<AppUpdateView />
<Header />
...
```
No check for `isDesktopView` needed. _I know we discussed the other way, but this looks better in UI._

src/renderer/i18n/de/update.ts
```diff
@@ -0,0 +1,9 @@
+import { UpdateMessages } from '../types'
+
+const update: UpdateMessages = {
+  'update.description': 'Neue Version {version} verfügbar',
+  'update.link': 'Release Webseite',
+  'update.checkFailed': 'Error while check for update: {error} - DE'
```
veado:
```
Error beim Update-Check: {error}
```

src/main/api/appUpdate.ts
```diff
@@ -0,0 +1,60 @@
...
+import { autoUpdater } from 'electron-updater'
+
+import { AppUpdateRD } from '../../shared/api/types'
+import IPCMessages from '../ipc/messages'
+
+autoUpdater.logger = {
+  ...log,
+  // Set custom info logger for better understanding in common debug-flow
+  info(message?: unknown): void {
+    console.log('[AutoUpdater.info]', message)
```
veado: ^ In `dev` mode only?

veado (changes requested):

dev-app-update.sample.yml
```diff
@@ -0,0 +1,5 @@
+provider: github
+owner: 'thorchain'
+repo: 'asgardex-electron'
+# @see https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token
+token: githubPersonalAccessToken
```
veado: [nitpick]
```
# GitHub personal access token
# @see https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token
token: XXX
```

src/main/api/appUpdate.ts
```diff
@@ -0,0 +1,62 @@
...
+        autoUpdater.removeAllListeners('update-not-available')
+      }
+
+      return new Promise((resolve) => {
+        autoUpdater.once('update-available', (info: { version: string }) => {
+          resolve(RD.success(info.version))
+        })
+
+        autoUpdater.once('update-not-available', () => {
+          resolve(RD.initial)
```
veado, line 38: It should be still `success` to handle a "Not available" status at `render` side. Also we do need to handle `pending` state (here in `main` or in `renderer` thread). Because as a next step we will let the user run a `re-check` (e.g. in case of `failure` or by providing a "check for updates" button in `Settings`). Not needed with this PR, just to let you know about it.

src/main/ipc/messages.ts
```diff
@@ -8,6 +8,9 @@ enum IPCMessages {
   LOAD_KEYSTORE = 'LOAD_KEYSTORE',
   GET_LEDGER_ADDRESS = 'GET_LEDGER_ADDRESS',
   SEND_LEDGER_TX = 'SEND_LEDGER_TX',
+  UPDATE_AVAILABLE = 'UPDATE_AVAILABLE',
+  UPDATE_NOT_AVAILABLE = 'UPDATE_NOT_AVAILABLE',
```
veado: Are both messages ^ still needed?

src/renderer/components/AppUpdate/AppUpdate.tsx
```diff
@@ -0,0 +1,75 @@
...
+    }
+  | {
+      isOpen: false
+    }
+
+export const AppUpdate: React.FC<AppUpdateModalProps> = (props) => {
+  const intl = useIntl()
+  const isDesktopView = Grid.useBreakpoint()?.lg ?? false
+
+  if (props.isOpen && props.type === 'success') {
```
veado, line 30: Pls destruct values from `props` before (as we do it in other React components).

src/renderer/components/AppUpdate/AppUpdate.tsx
```diff
@@ -0,0 +1,75 @@
...
+    }
+  | {
+      isOpen: true
+      type: 'fail'
+      close: () => void
+      message: string
+    }
+  | {
+      isOpen: false
+    }
```
veado, line 24: Why do we need union types for props here ^?

src/renderer/views/app/AppUpdateView.tsx
```diff
@@ -0,0 +1,33 @@
...
+      FP.pipe(
+        appUpdater,
+        RD.fold(
+          (): AppUpdateModalProps => ({ isOpen: false }),
+          (): AppUpdateModalProps => ({ isOpen: false }),
+          ({ message }): AppUpdateModalProps => ({ isOpen: true, type: 'fail', close: resetAppUpdater, message }),
+          (version): AppUpdateModalProps => ({
+            isOpen: true,
+            type: 'success',
+            goToUpdates: () => window.apiUrl.openExternal(`${packageInfo.repository.url}/releases/tag/v${version}`),
```
veado: [nitpick] Let's use `packageInfo.repository.url` in `ExternalUrl.GITHUB` (`src/shared/const.ts`). Or remove `ExternalUrl.GITHUB` there to use `packageInfo.repository.url` everywhere. _Just to avoid duplication ..._

veado (changes requested):

src/main/api/appUpdate.ts
thatStrangeGuyThorchain: yeah. i thought about `RD.success` but in this case we have to have `RD.RemoteData<Error, O.Option<string>>` to understand if there is "real" update available which i thought a little bit redundant. if this is just an `RD.RemoteData<Error, string>` we can just resolve Promise with `RD.initial` state to tell to the renderer process that there is no real update available. does it make sense? if no - i can re-factor with `RemoteData<Error, Option<string>>`

about `RD.pending`. it should be handled at the `useAppUpdate`-level (when requesting/re-requesting info about updates) as as  soon as we use promise (native way for electron) to communicate between `main` and `renderer` processes we can not send more than a single value so from `main` process we can send **only** a single value

veado (dismissed):

Issue #1390, opened 2021-04-30
[Swap] Test latest helpers of `Swap.utils`


veado:
- `minAmountToSwapMax1e8`
- `minBalanceToSwap`

^ All added by #1394

Issue #1381, opened 2021-04-27
ADD: Better Fees Handling


thorchain-admin:
# Desired

A simple rule of thumb to follow. 

This issue relies on implementing this: https://github.com/xchainjs/xchainjs-lib/issues/299 first, which is to change the suggested fees from thirdparty to THORChain gas oracle. 

## FEES
Three fees in THORChain:
1) The on-chain deposit transaction fee (inbound tx)
2) The liquidity fee as a function of slip
3) The on-chain redemption transaction fee (outbound tx)

> For clarity, the two chains will be referred to as the `inboundChain` and the `outboundChain`

### Inbound Fee

Get from `inboundChain.getFee(params, "fast")` -> the fee the user will pay to make the tx

### Liquidity Fee

Do not show to the user, since it is included in the quote

### Outbound Fee

Get from `3 * outboundChain.getFee(params, "fast")` -> the fee that will be deducted from outbound/refund/withdrawal

## THRESHOLDS

If a user swaps/adds liquidity to a chain, two things could happen:
1) swap is successful, the outboundFee will be deducted from the output
2) swap/add fails, the outboundFee will be deducted from the refund

So, do not let the user send an amount less than the higher of the two. 

## Implementation

Firstly, make sure the outboundFee for calculations is computed in the value of the inbound side:
```
outboundFeeInInboundAsset = getValueOfToken1InToken2(outboundFee, outboundAsset , inboundAsset)
```

"Estimated Fee". -> Display this to the user
`1.5 * minInputAmount` -> Do not let the user input less than this (use form validation). Over-estimate by 50% (1.5x) to prevent issues. 

### Swaps 

```
"Estimated Fee" == "inboundFee" + "outboundFee"
minInputAmount = 1.5 * max(inboundFee, outboundFeeInInboundAsset) // swap or refund
```

### Add liquidity 

```
"Estimated Fee" == "inboundFee" + "outboundFee"
minInputAmount = 1.5 * max(inboundFee, outboundFeeInInboundAsset)  // add or refund
```

### Withdraw Liquidity 

```
"Estimated Fee" == "inboundFee" + "outboundFee" // withdrawals
```

thorchain-admin:
A faster way to do it

## InboundFee
BTC/LTC/BCH = 1 * gasRate (sat/byte) * 250 (bytes) 
BNB = 1 * gasRate (sat/byte) * 1 (bytes) 
[...]

PR #1394, opened 2021-05-04
Better Swap fee handling (part 3): Update logic to check min. values

veado:
to cover all possible swap fees

- [x] Extract `poolFee$`  + `getChainFeeByGasRate` to remove `getInboundFee` / `getOutboundFee` (based on suggestion by @thatStrangeGuyThorchain)
- [x] Refactor `swapFees$` to use these changes ^
- [x] Add helpers `priceFeeAmountForInAsset` + `minAmountToSwapMax1e8` + `minBalanceToSwap` to `Swap.utils` (tests will be added with another PR - see #1390)
- [x] Refactor `Swap.tsx` to use these helpers ^
- [x] Update to latest `xchain-*` packages

P.S.: `minAmountToSwapMax1e8` includes all the calculation we discussed in our meeting today.

Part of #1381


src/renderer/components/swap/Swap.utils.ts
```diff
@@ -155,3 +165,117 @@ export const poolAssetDetailToAsset = (oAsset: O.Option<PoolAssetDetail>): O.Opt
...
+ * In other case use `minAmountToSwapMax1e8` to get min value
+ */
+export const minBalanceToSwap = (swapFees: SwapFees): BaseAmount => {
+  const {
+    inFee: { amount: inFeeAmount },
+    refundFee: { amount: refundFeeAmount }
+  } = swapFees
+
+  // Sum inbound (success swap) + refund fee (failure swap)
+  const feeToCover: BaseAmount = inFeeAmount.plus(refundFeeAmount)
```
thatStrangeGuyThorchain, line 278: are this BaseAmounts based on the same asset ? 🤔 
i mean if there any possibility that we will sum for example BNB-based fee and BUSD-based fee..?

veado: Both txs are running on same chain. For example swapping `BNB.BNB` -> `ETH.USDT`
- `inFee` (or `inboundFee`) in case of success is a tx into BNB pool. That's fee in BNB.
- `refundFee` in case of failure is a tx back from BNB pool to users BNB account. That's a BNB fee as well. So it does not matter what `outbound` asset or chain is (ETH chain in this case).

thatStrangeGuyThorchain (approved):

PR #1411, opened 2021-05-10
update deps 07 05 2021

thatStrangeGuyThorchain:
Also includes manual clean-up of `yarn.lock`

packages not updated:
```
ncu
Using yarn
Checking /Users/achepurko/Documents/projects/asgardex-electron/package.json
[====================] 104/104 100%

 chart.js           ^2.9.4  →   ^3.2.1     // should be separate PR as it needed to migrate from 2x to 3x version
 cosmos-client     ^0.39.2  →  ^0.42.4     // should be updated at the `xchain-cosmos` first
 react-chartjs-2   ^2.11.1  →   ^3.0.3     // same as `chart.js`
 rxjs               ^6.6.7  →   ^7.0.0     // can not upgrade rxjs to 7x version as there is a conflict with
                                           // @devexperts/rx-utils/dist/observable.utils (see screenshot)
 @storybook/react  ^6.1.21  →   ^6.2.9     // depends on babel loader which version is now not acceptable from CRA
 prettier           ^2.2.1  →   ^2.3.0     // will be updated with next PR as it causes large diff
 ts-loader          ^8.1.0  →   ^9.1.2
```
`ts-loader` can not update to 9x version as [it does not support webpack<5x](https://github.com/TypeStrong/ts-loader/releases/tag/v9.0.0) which is used by craco

image


package.json
```diff
@@ -136,7 +136,7 @@
     "@types/fs-extra": "^9.0.11",
     "@types/jest": "^26.0.23",
     "@types/lodash.shuffle": "^4.2.6",
-    "@types/node": "^14.14.37",
+    "@types/node": "^15.0.2",
```
veado: We don't use Node@15, let's stay with `14.x`

veado (changes requested):
From prettier blog about `Prettier 2.3.` https://prettier.io/blog/2021/05/09/2.3.0.html
> This release focuses on fixing long-standing issues in the JavaScript printer. Be warned that, unfortunately, reformatting a project with the new version might result in quite a big diff. If you don’t use ignoreRevsFile to hide such wholesale changes from git blame, it might be about time.

Pls try to revert changes made by latest prettier to follow suggestion for using `ignoreRevsFile` , Iit might reduce these massive changes in this PR (hopefully)

thatStrangeGuyThorchain:
> From prettier blog about `Prettier 2.3.` https://prettier.io/blog/2021/05/09/2.3.0.html
> 
> > This release focuses on fixing long-standing issues in the JavaScript printer. Be warned that, unfortunately, reformatting a project with the new version might result in quite a big diff. If you don’t use ignoreRevsFile to hide such wholesale changes from git blame, it might be about time.
> 
> Pls try to revert changes made by latest prettier to follow suggestion for using `ignoreRevsFile` , Iit might reduce these massive changes in this PR (hopefully)

as discussed at the DMs all prettier-related changes will be moved to the separate PR as [github does does not support such feature](https://github.community/t/support-ignore-revs-file-in-githubs-blame-view/3256/23). it's only for local `git blame` configs

veado (approved):
Good work!

PR #1418, opened 2021-05-11
[withdraw] Handle min value

veado:
- [x] Add `minAssetAmountToWithdrawMax1e8` + `minRuneAmountToWithdraw` to `Withdraw.helper`
- [ ] Add tests ^ (_will be another PR_)
- [x] Update `Withdraw` component to handle min. value, show fees for all txs
- [x] Update `Withdraw.stories`
- [x] Extract `priceFeeAmountForAsset` into `fees/utils` to remove similar helper from  from `Swap.util` / `Deposit.helper`
- [x] Update styles of `Slider` to visualize errors

Part of #1381

### DIFF
--- a/src/renderer/components/deposit/withdraw/Withdraw.tsx
+++ b/src/renderer/components/deposit/withdraw/Withdraw.tsx
@@ -34,6 +34,7 @@ import {
   SymWithdrawFeesRD,
   SymWithdrawFees
 } from '../../../services/chain/types'
+import { PoolsDataMap } from '../../../services/midgard/types'
 import { ValidatePasswordHandler } from '../../../services/wallet/types'
 import { AssetWithDecimal } from '../../../types/asgardex'
 import { PasswordModal } from '../../modal/password'
@@ -69,6 +70,7 @@ export type Props = {
   withdraw$: SymWithdrawStateHandler
   fees$: SymWithdrawFeesHandler
   network: Network
+  poolsData: PoolsDataMap
 }
 
 /**
@@ -91,13 +93,16 @@ export const Withdraw: React.FC<Props> = ({
   reloadFees,
   withdraw$,
   fees$,
-  network
+  network,
+  poolsData
 }) => {
   const intl = useIntl()
 
   const { asset, decimal: assetDecimal } = assetWD
 
   const [withdrawPercent, setWithdrawPercent] = useState(disabled ? 0 : 50)
+
+  const zeroWithdrawPercent = useMemo(() => withdrawPercent <= 0, [withdrawPercent])
 
   const {
     state: withdrawState,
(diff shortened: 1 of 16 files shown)


src/renderer/components/deposit/withdraw/Withdraw.tsx
```diff
@@ -313,42 +336,74 @@ export const Withdraw: React.FC<Props> = ({
...
-
-  const disabledForm = useMemo(() => withdrawPercent <= 0 || disabled, [withdrawPercent, disabled])
+    reloadFees(asset)
+  }, [reloadFees, asset])
+
+  // Load fees by every `onMount`
+  useEffect(() => {
+    reloadFees(asset)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
thatStrangeGuyThorchain, line 355: just to clarify and keep in mind. right now this works as we use flag `destroyInactiveTabPane` at the [Deposit.tsx component](https://github.com/thorchain/asgardex-electron/blob/develop/src/renderer/components/deposit/Deposit.tsx#L153). reloading deposit fees is working now for every switch between `Deposit` and `Withdraw` screens for the same reason. 
we just should keep in mind that if we decide (for any reason) to remove that `destroyInactiveTabPane` tab we have to reload fees for every chain's switching

src/renderer/services/chain/fees/common.ts
```diff
@@ -21,7 +20,7 @@ export const poolFee$ = (asset: Asset): PoolFeeLD => {
   if (isRuneNativeAsset(asset)) {
     return FP.pipe(
       THOR.fees$(),
-      liveData.map((fees) => ({ amount: fees[FeeOptionKeys.SWAP], asset: AssetRuneNative }))
+      liveData.map((fees) => ({ amount: fees['fast'], asset: AssetRuneNative }))
```
thatStrangeGuyThorchain: just a nitpick
```
liveData.map((fees) => ({ amount: fees.fast, asset: AssetRuneNative }))
```

thatStrangeGuyThorchain (dismissed):

src/renderer/components/deposit/withdraw/Withdraw.tsx
thatStrangeGuyThorchain: hm... in general... i'd suggest to keep depending on the `asset` value to trigger an effect to avoid such possible hard-to-understand situation. anyway we can not change an asset from `Withdraw` form...

veado: Good points. The reason for this `useEffect`is: If we don't reload fees ẃith `onMount`, previous fees are still visible. Since adding `[asset]` has no effect, I think it's ok to stay with current approach.

thatStrangeGuyThorchain: yes. correct. 

> Since adding [asset] has no effect, I think it's ok to stay with current approach.

yep. right now there is no difference. lets keep only `onMount` effect but also keep in mind such scenario

Issue #1305, opened 2021-04-12
UX Improvements


thorchain-admin:
# Pool Rows
## 1 _~~Green Selected Line Under Pool Type~~_ #1466 
<img width="369" alt="image" src="https://user-images.githubusercontent.com/41812729/114383390-d3921080-9bd0-11eb-80e5-57352be268c8.png">

## ~~_2 Default Sort is Depth_~~ **DONE**
<img width="394" alt="image" src="https://user-images.githubusercontent.com/41812729/114383443-e573b380-9bd0-11eb-8103-dad34e208ed6.png">

## _~~3  Persist Sort Selection After Navigation~~_ **Duplicated by #1187**

# Pool Details
## _~~4 Logo to Left of Asset On Detail View~~_ #1467  
Put the asset logo just to the left of the asset name, same size as the Pool Row Icon
<img width="300" alt="image" src="https://user-images.githubusercontent.com/41812729/114383609-148a2500-9bd1-11eb-9511-00f1217b1179.png">

## 5 _~~Reduce Vertical Height of Tx Rows~~_ #1468 
Row Height (and pill height) is too high, make 50% height to squeeze more rows in. 

<img width="920" alt="image" src="https://user-images.githubusercontent.com/41812729/114383652-2075e700-9bd1-11eb-897b-cd3b205da166.png">

## ~~Alignment Issue~~ **Fixed by #1473**

<img width="362" alt="image" src="https://user-images.githubusercontent.com/41812729/114383855-5b781a80-9bd1-11eb-8ffa-a259983eda91.png">

# ~~Swap Page~~

**Not fixable easily (since it's default behavior of `antd` slider)** _We have decided just to ignore it..._

## Horizontal Alignment

<img width="283" alt="image" src="https://user-images.githubusercontent.com/41812729/114383996-85c9d800-9bd1-11eb-8245-ef24313fcf5f.png">

# _~~Tx Confirmation Modal~~_ #1469

## _~~Lower case `s` instead of `S` for the seconds~~_
<img width="224" alt="image" src="https://user-images.githubusercontent.com/41812729/114384097-a2fea680-9bd1-11eb-8ee1-eb486587c99c.png">

## _~~Remove the `0.00%` green slip field~~_

<img width="426" alt="image" src="https://user-images.githubusercontent.com/41812729/114392150-b878ce00-9bdb-11eb-9fa9-204117d22180.png">

# Add Liquidity

## ~~_Balance Of Asset done_~~#1471

[...]

PR #1466, opened 2021-05-27
 Green Selected Line Under Pool Type

thatStrangeGuyThorchain:
- added underline after active filter

as a part of #1305


veado (approved):
LGTM

PR #1468, opened 2021-05-27
Reduce Vertical Height of Tx Rows

thatStrangeGuyThorchain:
- decreased rows height
- changed `DEFAULT_PAGE_SIZE` to `10` for `PoolActionsHistory`

as a part of #1305


veado (approved):
LGTM

Issue #1373, opened 2021-04-25
Show 24hrVolume


veado:
at header (or at any other prominent place)

Formula: `totalVolume * runePriceUSD`. Get data from https://midgard.thorchain.info/v2/history/swaps?interval=day&count=1)

In future data will be available via https://midgard.thorchain.info/v2/stats (see [internal Slack thread](https://thorchain.slack.com/archives/CL5B4M4BC/p1619349238145800?thread_ts=1619349103.145200&cid=CL5B4M4BC))

Issue #1328, opened 2021-04-14
ADD: RUNE Price


thorchain-admin:
# Add RUNE Price on Header
Right next to the Da/Night Icon add `RUNE: $14.50` with grey pill shading. 

Get price from here: https://midgard.thorchain.info/v2/stats

<img width="497" alt="image" src="https://user-images.githubusercontent.com/41812729/114664705-6cdf3500-9d3f-11eb-8843-ae3747db25c1.png">

<img width="505" alt="image" src="https://user-images.githubusercontent.com/41812729/114664712-71a3e900-9d3f-11eb-8ac1-e669251250f2.png">

PR #1472, opened 2021-05-28
Update header UI to show RUNE price / 24h volume

veado:
- [x] Introduce `HeaderStats` to show RUNE price + 24hrs volume in header

Note: Business logic will be added with another PR

Peek 2021-05-28 18-03

Part of #1373, #1328

### DIFF
--- a/src/renderer/components/header/HeaderComponent.tsx
+++ b/src/renderer/components/header/HeaderComponent.tsx
@@ -1,5 +1,6 @@
 import React, { useMemo, useState, useCallback, useRef } from 'react'
 
+import { assetAmount, assetToBase } from '@xchainjs/xchain-util'
 import { Row, Col, Tabs, Grid } from 'antd'
 import * as FP from 'fp-ts/function'
 import * as A from 'fp-ts/lib/Array'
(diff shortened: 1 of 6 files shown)


src/renderer/components/header/HeaderComponent.tsx
```diff
@@ -300,6 +303,15 @@ export const HeaderComponent: React.FC<Props> = (props): JSX.Element => {
                   <AsgardexLogo />
                   {renderHeaderNetStatus}
                   <HeaderTheme isDesktopView={isDesktopView} />
+                  {isLargeDesktopView && (
```
veado, line 306: Show price + volume for large desktops only. In future we will clean up the header to move some settings into `Settings` (e.g. language, theme), but not for this or next week :)

src/renderer/components/header/HeaderComponent.tsx
```diff
@@ -300,6 +303,15 @@ export const HeaderComponent: React.FC<Props> = (props): JSX.Element => {
...
+                    <HeaderStats
+                      runePrice={assetAmount(14.08)}
+                      volume24={{
+                        asset: AssetBUSDBD1,
+                        amount: assetToBase(assetAmount('24000000'))
+                      }}
+                    />
+                  )}
```
thatStrangeGuyThorchain, line 314: should this values be hardcoded ? 🤔

thatStrangeGuyThorchain (changes requested):

thatStrangeGuyThorchain (approved):

PR #1483, opened 2021-06-01
Fix loading state for RUNE price in header

### LINKED ISSUE #1482: [Header] RUNE price shows wrong value by re-loading data
Peek 2021-06-01 11-01

veado:
- [x] Fix loading state for RUNE price in header
- [x] Introduce `useRunePrice` hook

Peek 2021-06-01 10-54

Fix #1482

### DIFF
--- a/src/renderer/components/header/stats/HeaderStats.tsx
+++ b/src/renderer/components/header/stats/HeaderStats.tsx
@@ -45,7 +45,7 @@ export const HeaderStats: React.FC<Props> = (props): JSX.Element => {
               prevRunePriceLabel.current = label
               return label
             }),
-            O.getOrElse(() => '')
+            O.getOrElse(() => '--')
           )
         )
       ),
--- a/src/renderer/components/header/stats/HeaderStats.tsx
+++ b/src/renderer/components/header/stats/HeaderStats.tsx
@@ -1,5 +1,6 @@
 import React, { useMemo, useRef } from 'react'
 
+import * as RD from '@devexperts/remote-data-ts'
 import { baseToAsset, formatAssetAmountCurrency } from '@xchainjs/xchain-util'
 import * as FP from 'fp-ts/lib/function'
 import * as O from 'fp-ts/lib/Option'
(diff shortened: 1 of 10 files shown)


src/renderer/components/header/stats/HeaderStats.tsx
```diff
@@ -1,54 +1,63 @@
...
+                trimZeros: true,
+                decimal: isUSDAsset(asset) ? 2 : 6
+              })
+            ),
+            O.map((label) => {
+              // store price label
+              prevRunePriceLabel.current = label
+              return label
+            }),
+            O.getOrElse(() => '')
```
veado: In case of no price, it should be `--` ^

src/renderer/hooks/useRunePrice.types.ts
```diff
@@ -0,0 +1,6 @@
+import * as RD from '@devexperts/remote-data-ts'
+import * as O from 'fp-ts/lib/Option'
+
+import { AssetWithAmount } from '../types/asgardex'
+
+export type RunePriceRD = RD.RemoteData<Error, O.Option<AssetWithAmount>>
```
thatStrangeGuyThorchain: not sure that i understand why we need option inside ?

src/renderer/views/pools/Pools.types.ts
```diff
@@ -22,6 +24,9 @@ export type PricePool = {
   poolData: PoolData
 }
 
+export type PricePool$ = Rx.Observable<PricePool>
+export type PricePoolRD = RD.RemoteData<Error, PricePool>
```
thatStrangeGuyThorchain: hm. this time i could not fins usage of `PricePoolRD` 😅

src/renderer/hooks/useRunePrice.types.ts
veado: Price might not be available in case there is no price pool or there is no selected price asset.

src/renderer/views/pools/Pools.types.ts
veado: Yeah, that happens if you tried to solve that issue, but using a different way before :sweat_smile:

src/renderer/hooks/useRunePrice.types.ts
thatStrangeGuyThorchain: yep. but is not `RD.initial` enough in this case ?
- `initial` : no data provided
- `pending`: loading data
- `error`: error data fetching
- `success`: show data
?

veado: ^ solved by @thatStrangeGuyThorchain :eye:  a1c50638f119bdb30c5ff6f81adef3cfd9c5a5f3

PR #1485, opened 2021-06-01
[Header] Reload volume24 / rune price by click

veado:
https://user-images.githubusercontent.com/61792675/120361693-6e25ea80-c30a-11eb-8dcd-d607b4d07c8d.mp4

### DIFF
--- a/src/renderer/hooks/useVolume24Price.ts
+++ b/src/renderer/hooks/useVolume24Price.ts
@@ -8,26 +8,42 @@ import * as RxOp from 'rxjs/operators'
 
 import { useMidgardContext } from '../contexts/MidgardContext'
 import { sequenceTRD } from '../helpers/fpHelpers'
+import { triggerStream } from '../helpers/stateHelper'
 import { PriceRD } from '../services/midgard/types'
 import { AssetWithAmount } from '../types/asgardex'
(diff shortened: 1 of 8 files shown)


src/renderer/hooks/useVolume24Price.ts
```diff
@@ -8,26 +8,42 @@ import * as RxOp from 'rxjs/operators'
...
+    FP.pipe(
+      reloadHistory$,
+      RxOp.switchMap((_) => apiGetLiquidityHistory$({ interval: GetLiquidityHistoryIntervalEnum.Day, count: 1 }))
+    )
+
+  const reloadVolume24Price = () => {
+    // reload of pools are needed to calculate prices properly
+    reloadPools()
+    reloadHistory()
+  }
```
thatStrangeGuyThorchain, line 41: dont we want to use `useMemo/useCallback`'s ?

thatStrangeGuyThorchain (approved):

src/renderer/hooks/useVolume24Price.ts
veado: I think it's not needed here, all are just lazy functions without any "heavy" computations.

PR #1511, opened 2021-06-07
[Resolve]: [Withdraw] Min. amount not visible (dark theme only) 

### LINKED ISSUE #1506: [Withdraw] Min. amount not visible (dark theme only)
`min` amount not visible (wrong color) using dark theme

Peek 2021-06-03 20-51

thatStrangeGuyThorchain:
-  replaced using plain `span` elements to `UILabel`

closes #1506


veado (approved):
:cool:

PR #1524, opened 2021-06-09
Resolve: "Get rid of chaosnet in type Network in favour of using mainnet only"

### LINKED ISSUE #1522: Get rid of `chaosnet` in type `Network` in favour using `mainnet` only
### HAVE
```ts
export type Network = 'testnet' | 'chaosnet' | 'mainnet'
```

### WANT
```ts
export type Network = 'testnet' | 'mainnet'
```

veado:
Closes #1522

### DIFF
--- a/src/shared/api/types.ts
+++ b/src/shared/api/types.ts
@@ -65,7 +65,7 @@ export type ApiUrl = {
   openExternal: (url: string) => Promise<void>
 }
 
-export type Network = 'testnet' | 'chaosnet' | 'mainnet'
+export type Network = 'testnet' | 'mainnet'
 
 export enum LedgerErrorId {
   NO_DEVICE,
--- a/src/shared/const.ts
+++ b/src/shared/const.ts
@@ -1,5 +1,3 @@
-import * as A from 'fp-ts/lib/Array'
-
 import { envOrDefault } from '../renderer/helpers/envHelper'
 import { StoreFilesContent, UserNodesStorage } from './api/types'
 import { Locale } from './i18n/types'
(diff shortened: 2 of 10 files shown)


src/shared/api/types.ts
```diff
@@ -65,7 +65,7 @@ export type ApiUrl = {
   openExternal: (url: string) => Promise<void>
 }
 
-export type Network = 'testnet' | 'chaosnet' | 'mainnet'
+export type Network = 'testnet' | 'mainnet'
```
veado, line 68: Note: Same type as we have with `Network` in `xchain-client`. However, I would still stick with a custom type here to be independent of `xchain-*` - just in case `xchain-*` will change it for whatever reason in the future.

src/shared/const.ts
```diff
@@ -21,9 +19,8 @@ const USER_NODES_STORAGE_VERSION = '1'
 
 export const USER_NODES_STORAGE_DEFAULT: UserNodesStorage = {
   version: USER_NODES_STORAGE_VERSION,
-  mainnet: A.empty,
-  chaosnet: A.empty,
-  testnet: A.empty
+  mainnet: [],
```
veado, line 22: @thatStrangeGuyThorchain BTW: `fp-ts` has marked `A.empty` as deprecated with a note `Use a new [] instead.`, but I could never find such "new" type `[]` there. Any idea?

thatStrangeGuyThorchain: not sure but i suppose that means exactly what you have done:
create a new empty array... 
actually seems a pointer to a single instance of an array is unsafe actually 'cuz anyone can push or smth like that to it. i mean this is a mutable object and it seems this is a reason why `fp-ts` deprecated it.

another safely way to use empty array from `fp-ts` as initial value here is to use `ReadOnlyArray`
so with this it will be like this:
```ts
import * as ROA from 'fp-ts/ReadOnlyArray`

export const USER_NODES_STORAGE_DEFAULT: UserNodesStorage = {
  version: USER_NODES_STORAGE_VERSION,
  mainnet: ROA.empty,
  testnet: ROA.empty
}
```
which is `read-only` and safer at type-/runtime-levels

thatStrangeGuyThorchain (approved):

PR #1640, opened 2021-07-18
Feature/update fr translation

Weyland2093:
french translation updated


veado:
@Weyland2093 Thanks :+1:

PR #1641, opened 2021-07-20
Disable swap on slippage

### LINKED ISSUE #1105: [Swap] Slip limit exceeded check
Right now, we don't have slip limit exceeded check in swap page.
If the user send tokens to swap even though the slip limit is exceeded, then It is refunded so user lose the fees.
So in this case, the drag button needs to be disabled and show proper msg.

image

gromxyz:
- Slip percentage is shown
- if it's higher than slip tolerance it gets red and disables swap button

Closes #1105

### DIFF
--- a/src/renderer/components/swap/Swap.tsx
+++ b/src/renderer/components/swap/Swap.tsx
@@ -275,10 +275,7 @@ export const Swap = ({
     return max1e8BaseAmount(swapResultAmount)
   }, [swapData.swapResult, targetAssetDecimal])
 
-  const isCausedSlippage = useMemo(() => {
-    const limitFromAmount = amountToSwapMax1e8.amount().multipliedBy((100 - slipTolerance) / 100)
-    return amountToSwapMax1e8.div(bn(100)).times(swapData.slip).gt(amountToSwapMax1e8.minus(limitFromAmount))
-  }, [amountToSwapMax1e8, swapData.slip, slipTolerance])
+  const isCausedSlippage = useMemo(() => swapData.slip > bn(slipTolerance), [swapData.slip, slipTolerance])
 
   const oApproveParams: O.Option<ApproveParams> = useMemo(() => {
     const oRouterAddress: O.Option<Address> = FP.pipe(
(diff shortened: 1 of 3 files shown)


src/renderer/components/swap/Swap.tsx
```diff
@@ -276,6 +275,11 @@ export const Swap = ({
     return max1e8BaseAmount(swapResultAmount)
   }, [swapData.swapResult, targetAssetDecimal])
 
+  const isCausedSlippage = useMemo(() => {
+    const limitFromAmount = amountToSwapMax1e8.amount().multipliedBy((100 - slipTolerance) / 100)
+    return amountToSwapMax1e8.div(bn(100)).times(swapData.slip).gt(amountToSwapMax1e8.minus(limitFromAmount))
+  }, [amountToSwapMax1e8, swapData.slip, slipTolerance])
```
veado: Could you extract this function into `Swap.utils` and add a test, pls?

veado (changes requested):
Pls show an error message to the user if slippage tolerance has been reached

veado:
Some tweaks needed:

- ~~Label of selected slippage tolerance has not been updated - it sticks with 0.00%~~
- Show "hand" courser (pointer) by hovering items select box
- Make gear icon small (check layout at Figma)
- Update color by hovering gear icon to "dark" grey

https://user-images.githubusercontent.com/61792675/126304092-76ac1254-8e58-4c84-b7b6-0281b73638e1.mp4

src/renderer/components/swap/Swap.tsx
gromxyz: Now it's simplified :)

veado:
Last (small) change requests:

- Fix `React` warning such as `React does not recognize the `isCausedSlippage` prop on a DOM element.` or `Received `false` for a non-boolean attribute 'active'` 

Screenshot from 2021-08-02 13-12-49
Screenshot from 2021-08-02 13-12-36

- Show mouse cursor while hovering gear icon. Increase area to react a mouse hover.


https://user-images.githubusercontent.com/61792675/127853601-ce928391-68f9-445b-94ae-648aeb1b7a61.mp4

PR #1648, opened 2021-08-03
Add swap limit protection

### LINKED ISSUE #1647: [Swap] Add swap limit protection to `swap` memo
## Description

We need to add limit to cover the case of unexpected slippage.
Limit sets the protection of minimum amount of destination asset to receive, otherwise it gets source asset back refunded by throchain network. 
`SWAP:{ASSET}:{DESTADDR}:{LIMIT}`

Calculation for the limit: `destination asset` * (1 - `slippage tolerance` * 0.01)

Example for 3% slip:
1 - 0.03 = 0.97
=> 97% minimum amount of the destination asset to receive

gromxyz:
- calculation of limit based on slippage tolerance
- added limit to swap memo

Closes #1647


veado (approved):
LGTM

PR #1676, opened 2021-08-16
[Settings] Enable access to Settings without imported wallet

### LINKED ISSUE #1575: [Settings] Extract wallet content to `Wallet` section
1. In Wallet section: Add a fifth tab `Settings`
2. Move content of `Account management` + `Wallet Management` from `Settings` to `Wallet / Settings`
3. Update routes to let user access `app` settings w/o restriction. To access `Wallet/Settings` an user needs to have a wallet imported and unlocked (Check `WalletView` to see current implementation for such routes)
4. Re-organize some elements to fill empty space

gromxyz:
- [x] Enable access to Settings without imported wallet

Closes #1575

### DIFF
--- a/src/renderer/views/wallet/WalletView.tsx
+++ b/src/renderer/views/wallet/WalletView.tsx
@@ -215,9 +215,6 @@ export const WalletView: React.FC = (): JSX.Element => {
           <Route path={walletRoutes.base.template} exact>
             <Redirect to={walletRoutes.assets.path()} />
           </Route>
-          <Route path={walletRoutes.settings.template} exact>
-            <SettingsView />
-          </Route>
           <Route path={walletRoutes.assets.template} exact>
             {reloadButton(reloadBalances)}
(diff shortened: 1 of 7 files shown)


src/renderer/views/wallet/WalletView.tsx
```diff
@@ -343,6 +340,9 @@ export const WalletView: React.FC = (): JSX.Element => {
       <Route path={walletRoutes.imports.base.template}>
         <ImportsView />
       </Route>
+      <Route path={walletRoutes.settings.template} exact>
+        <SettingsView />
+      </Route>
```
veado: Since `Settings` are global or app settings now, let's extract that route from `walletRoutes` into `appRoutes` using a path `/settings`. With that, route handling for `<SettingsView />` can be extracted into `<AppView />`

veado (changes requested):

veado (approved):

PR #1688, opened 2021-08-20
Allow adding custom swap recipient

### LINKED ISSUE #1683: [Swap] Add input field to enter recipient
## TODO

Create a custom component with following features:

- 3 States
   - DEFAULT
      - Show keystore wallet address (shorten) by default
      - Add icons for edit, copy, external link
   - EDIT
      - Show input field with full address to edit
      - Add icons to confirm or cancel
      - Error check while typing (see SendFormXYZ components how we deal validation)
   - ERROR
      - Show error in case of invalid address
 

### Swap page
Screenshot from 2021-08-18 14-20-08
 ^ https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020?node-id=0%3A1

### Input states
Screenshot from 2021-08-18 14-19-45
^  https://www.figma.com/file/LgRqZCGMShZ2mJ702MIqyV/ASGARDEX-2020?node-id=0%3A1

gromxyz:
- [x] create prototype on `Swap`
- [x] extract to `CustomAddressInput` to separate component
- [x] separate to `DEFAULT` and `EDITABLE` states
- [x] add i18n
- [x] enable editing
- [x] enable copy button
- [x] get target chain from asset to provide for opening externally
- [x] enable open external
- [x] styling adjustments
- [x] adjust state management
- [x] extract props
- [x] add types
- [x] add story
- [x] add tests
- [x] extract address link handler into hook
- [x] address validation

Closes #1683


src/renderer/components/swap/CustomAddressInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+import { useValidateAddress } from '../../hooks/useValidateAddress'
+import { AddressValidation } from '../../services/clients/types'
+import * as Styled from './CustomAddressInput.styles'
+
+type Props = {
+  oTargetAsset: O.Option<Asset>
+  oTargetWalletAddress: O.Option<string>
+  clickAddressLinkHandler: (chain: Chain, address: Address) => void
+}
```
veado: As discussed in DM (Discord): Encapsulate this component as much as possible, so it can be used everywhere. The component has no knowledge about its parent.
```ts
type Props = {
  asset: Asset // maybe we don't need this if we provide `addressValidator`?
  address: Address
  onClickOpenAddress: (address: Address) => void // callback for clicking "open external" icon
  onChangeAddress: (address: Address) => void // whenever address has been changed and validated by clicking "save" icon, this callback will be called
  addressValidator: (address: Address) => boolean // validator function to check address
}
```

src/renderer/components/swap/CustomAddressInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+        oTargetAsset,
+        O.map((asset) => asset.chain),
+        O.getOrElse(() => THORChain)
+      ),
+    [oTargetAsset]
+  )
+
+  const [editModeActive, setEditModeActive] = useState(false)
+
+  const [recipientAddress, setRecipientAddress] = useState('')
```
veado: Rename `recipientAddress` -> `address` (the component has no knowledge this address is a `Recipient`, it can be any address)

src/renderer/components/swap/CustomAddressInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+        O.map((address) => {
+          setRecipientAddress(address)
+          return address
+        }),
+        O.getOrElse(() => '')
+      ),
+    [oTargetWalletAddress]
+  )
+
+  const addressValidation: AddressValidation = useValidateAddress(BNBChain)
```
veado: Pass a validator using props. `useValidateAddress` should not be used in a component, since it has access to `ChainContext` (Remember: Any `XYZContext` or `useContextXYZ` should be used in `Views` only)

src/renderer/components/swap/CustomAddressInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+        O.map((asset) => asset.chain),
+        O.getOrElse(() => THORChain)
+      ),
+    [oTargetAsset]
+  )
+
+  const [editModeActive, setEditModeActive] = useState(false)
+
+  const [recipientAddress, setRecipientAddress] = useState('')
+  const [editableRecipientAddress, setEditableRecipientAddress] = useState('')
```
veado:
```ts
// O.some => editable = true
// O.none => editable = false
// No need to have `editModeActive` anymore
const [editableAddress, setEditableAddress] = useState<O.Option<Address>>(O.none)
```

src/renderer/components/swap/CustomAddressInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+        </Styled.AddressCustomRecipient>
+      </>
+    )
+  }
+
+  const renderEditable = () => {
+    return (
+      <Styled.EditableFormWrapper>
+        <Input value={editableRecipientAddress} onChange={(e) => setEditableRecipientAddress(e.target.value)} />
+        <Styled.ConfirmEdit style={{ margin: 5 }} onClick={saveCustomAddress} />
```
veado: [nitpick] If you have already defined a `Styled.XYZ`, then no need to use `style=...`. Put all styles in external `Styled.XYZ` component

src/renderer/components/swap/CustomAddressInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <Styled.CancelEdit style={{ margin: 5 }} onClick={cancelEditCustomAddress} />
```
veado: [nitpick] Extract `style=...` into `Styled.CancelEdit`

veado (changes requested):
Pls add a story for `CustomAddressInput` to test its behavior

src/renderer/components/swap/CustomAddressInput.tsx
gromxyz: Just wanted to make it work temporarily, will move it to the views level soon

gromxyz: That's also a reason why `BNBChain` is actually hardcoded here

src/renderer/hooks/useValidateAddress.ts
```diff
@@ -1,26 +1,41 @@
...
+
+  // `chainUpdated` needs to be called whenever chain has been updated
+  // to trigger `useObservableState` properly to get a client depending on chain
+  useEffect(() => chainUpdated(chain), [chain, chainUpdated])
 
   const validateAddress = useCallback(
-    (address: Address) =>
-      FP.pipe(
+    (address: Address) => {
+      return FP.pipe(
```
veado: `return` can be removed - it's just an artifact for debugging `address` before

src/renderer/components/swap/EditableAddress.tsx
```diff
@@ -0,0 +1,104 @@
...
+      if (!addressValidator(value)) {
+        return Promise.reject(intl.formatMessage({ id: 'wallet.errors.address.invalid' }))
+      }
+    },
+    [addressValidator, intl]
+  )
+
+  const [form] = Form.useForm<{ recipient: string }>()
+
+  const renderCustomAddress = useCallback(() => {
```
veado: `useCallback` -> `useMemo`

veado: [nitpick]  Rename `renderCustomAddress` -> `renderAddress`

src/renderer/components/swap/EditableAddress.tsx
```diff
@@ -0,0 +1,104 @@
...
+export const EditableAddress = ({
+  address,
+  onChangeAddress,
+  onClickOpenAddress,
+  addressValidator
+}: EditableAddressProps) => {
+  const RECIPIENT_FIELD = 'recipient'
+  const intl = useIntl()
+  const [editableAddress, setEditableAddress] = useState<O.Option<Address>>(O.none)
+  const maskedRecipientAddress = useMemo(() => address.substring(0, 7) + '...' + address.slice(-3), [address])
```
veado: There is already a component to shorten an address: `AddressEllipsis` It has a `ellipse` helper included. But maybe you just need to use `truncateAddress` from `helpers/AddressHelper` here ^

src/renderer/components/swap/EditableAddress.tsx
```diff
@@ -0,0 +1,104 @@
...
+                <Input />
+              </Form.Item>
+            </InnerForm>
+          </Styled.AddressFormWrapper>
+          <Styled.AddressEditButtonsWrapper>
+            <Styled.ConfirmEdit
+              onClick={() => {
+                if (form.getFieldError(RECIPIENT_FIELD).length === 0) {
+                  onChangeAddress(form.getFieldValue(RECIPIENT_FIELD))
+                  setEditableAddress(O.fromNullable(null))
```
veado: `setEditableAddress(O.fromNullable(null))` -> `setEditableAddress(O.none)`

src/renderer/components/swap/EditableAddress.tsx
```diff
@@ -0,0 +1,104 @@
...
+                }
+              }}
+            />
+            <Styled.CancelEdit onClick={() => setEditableAddress(O.fromNullable(null))} />
```
veado: `setEditableAddress(O.none)`

src/renderer/components/swap/EditableAddress.tsx
```diff
@@ -0,0 +1,104 @@
...
+      )
+    },
+    [form, onChangeAddress, validateAddress]
+  )
+
+  const renderCustomAddressInput = useMemo(
+    () =>
+      FP.pipe(
+        editableAddress,
+        O.map((editableAddress) => renderEditableCustomAddress(editableAddress)),
```
veado: [nitpick] ^ Can be shortened: `O.map(renderEditableCustomAddress),`

src/renderer/components/swap/EditableAddress.tsx
```diff
@@ -0,0 +1,104 @@
...
+        O.getOrElse(() => renderCustomAddress())
```
veado: In some cases it might be more readable to use `O.fold` (kinda pattern matching) instead of `O.map ... O.getOrElse`

```ts
FP.pipe(
  editableAddress,
  O.fold(
    // in case of `none`
    () => renderCustomAddress(),
    // in case of `some(...)`
    renderEditableCustomAddress // similar to ((editableAddress) => renderEditableCustomAddress(editableAddress)
  )   

```

src/renderer/components/swap/Swap.tsx
```diff
@@ -110,6 +111,8 @@ export type SwapProps = {
   importWalletHandler: FP.Lazy<void>
   haltedChains: Chain[]
   mimirHalt: MimirHalt
+  clickAddressLinkHandler: (address: Address) => void
+  addressValidator: (address: Address) => boolean
```
veado: `addressValidator: AddressValidation`

src/renderer/components/swap/Swap.tsx
```diff
@@ -1121,6 +1128,28 @@ export const Swap = ({
...
+      FP.pipe(
+        sequenceTOption(oTargetAsset, targetWalletAddress),
+        O.map(([asset, address]) => (
+          <EditableAddress
+            key={address}
+            asset={asset}
+            address={address}
+            onClickOpenAddress={(address) => clickAddressLinkHandler(address)}
+            onChangeAddress={(newAddress) => {
+              console.log(newAddress)
```
veado: Remove `console` statement

```ts
onChangeAddress={(newAddress) => setTargetWalletAddress(O.some(newAddress))
```

src/renderer/components/swap/Swap.tsx
```diff
@@ -1121,6 +1128,28 @@ export const Swap = ({
     [checkIsApprovedError, isApproveFeeError, walletBalancesLoading]
   )
 
+  const renderCustomAddressInput = useMemo(
+    () =>
+      FP.pipe(
+        sequenceTOption(oTargetAsset, targetWalletAddress),
+        O.map(([asset, address]) => (
```
veado: [nitpick] Maybe `O.fold` instead of `O.map ... O.getOrElse`?

src/renderer/i18n/de/swap.ts
```diff
@@ -6,6 +6,7 @@ const swap: SwapMessages = {
   'swap.state.success': 'Erfolgreich getauscht',
   'swap.input': 'Eingabe',
   'swap.output': 'Ausgabe',
+  'swap.recipient': 'Recipient - DE',
```
veado:
```ts
'swap.recipient': 'Empfänger'
```

src/renderer/components/swap/EditableAddress.styles.tsx
```diff
@@ -0,0 +1,100 @@
...
+export const ConfirmEdit = styled(CheckCircleOutlined)`
+  margin: 5px;
+  height: ${ICON_SIZE}px;
+  width: ${ICON_SIZE}px;
+  color: ${palette('primary', 0)};
+  svg {
+    color: inherit;
+    height: ${ICON_SIZE}px;
+    width: ${ICON_SIZE}px;
+  }
```
veado: To avoid duplications:

```ts
import styled, { css } from 'styled-components'

...

const iconStyles = css`
  margin: 5px;
  height: ${ICON_SIZE}px;
  width: ${ICON_SIZE}px;
  color: ${palette('primary', 0)};
  svg {
    color: inherit;
    height: ${ICON_SIZE}px;
    width: ${ICON_SIZE}px;
  }
`

export const ConfirmEdit = styled(CheckCircleOutlined)`
  ${iconStyles}
`
export const CancelEdit = styled(CloseCircleOutlined)`
  ${iconStyles}
`
// ^ same for other icons
```

veado (changes requested):

src/renderer/components/swap/EditableAddress.tsx
gromxyz: `AddressEllipsis` is not reusable in this case because of the positioning of the icons, since the copy would be the right to the address and then we lose control over the order of the icons.

The problem with `truncateAddress` is that I needs to know about the chain and network, which doesn't seem to me worth passing around through props, since existing solution could do just fine.

veado: `truncateAddress` is not as same as maskedRecipientAddress. `truncateAddress` is more "safety" by using a same pattern we used at other places. `chain` is already available in `asset` prop, so just a need to pass `network` down to `EditableAddress`.

src/renderer/components/swap/EditableAddress.styles.tsx
```diff
@@ -75,7 +75,7 @@ export const CancelEdit = styled(CloseCircleOutlined)`
   }
 `
 
-export const AddressLinkIcon = styled(ExternalLinkIconUI)`
+const iconStyles = css`
```
veado: You can use it ^ for `ConfirmEdit` + `CancelEdit` as well _I think_.

veado (changes requested):

veado:
Pls fix background color of the input (dark theme). Check `SendXYZForms` to get color values.
Screenshot from 2021-08-24 13-28-31

veado:
Pls disable `Swap` button whenever `EditableAddress` has an `editable` mode. That's needed to make sure an user has entered + saved a new address before swapping.

Screenshot from 2021-08-24 13-52-50

veado:
Mobile view needs to be tweaked a little bit - recipient address can be in same position as we have in desktop mode
Screenshot from 2021-08-24 13-57-34

~~Also, let's show shorten address in `mobile` mode only. In `desktop` mode there is good space to show full address.~~

veado (approved):

PR #1700, opened 2021-08-26
Resolve: [Wallet] Tag fees in txs history

gromxyz:
- [x] Add `FEE` in txs history
- [x] Quick fix: Typo in upgrade alert message
- [x] Add Ledger icon (committed by accident lol)

Closes #1698


veado (changes requested):
Make sure FEE is shown for **to** addresses only.

Pls check styles: `SELF` vs. `FEE` (it should be as same as `SELF`)

https://user-images.githubusercontent.com/61792675/130954952-341bd468-24c9-4657-9e5c-48bf0de99d62.mp4

src/renderer/components/wallet/txs/table/TxsTable.tsx
```diff
@@ -35,8 +35,9 @@ export const TxsTable: React.FC<Props> = (props): JSX.Element => {
     changePaginationHandler,
     network,
     chain,
-    walletAddress: oWalletAddres = O.none
+    walletAddress: oWalletAddress = O.none
   } = props
+  const FEE_RESERVE_ADDRESS = 'thor1dheycdevq39qlkxs2a6wuuzyn4aqxhve4qxtxt'
```
veado: Let's rename it to `RESERVE_MODULE_ADDRESS` and extract it into `src/renderer/services/thorchain/const.ts`

veado (changes requested):

veado (approved):

PR #1716, opened 2021-09-01
[Wallet] Get BNB Ledger address in settings

### LINKED ISSUE #1711: [Wallet] Get BNB Ledger address in settings
Code will be very similar to #1665, but for BNB

Part of #1709

gromxyz:
Closes #1711


src/main/api/ledger/binance.ts
```diff
@@ -1,32 +1,34 @@
...
-//   }
-// }
+export const getAddress = async (transport: Transport, network: Network) => {
+  try {
+    const app = new AppBNB(transport)
+    const derive_path = getDerivePath(0)
+    const { pk } = await app.getPublicKey(derive_path)
+    if (pk) {
+      // get address from pubkey
+      const address = crypto.getAddressFromPublicKey(pk.toString('hex'), network === 'testnet' ? 'tbnb' : 'bnb')
```
veado:
```
network === 'testnet' ? 'tbnb' : 'bnb'
```
^ Use `getPrefix` provided by `xchain-binance` https://github.com/xchainjs/xchainjs-lib/blob/master/packages/xchain-binance/src/util.ts#L131

veado (dismissed):
Looks good! Just a nitpick

src/main/api/ledger/binance.ts
```diff
@@ -1,32 +1,34 @@
...
+      return E.right(address)
+    } else {
+      return E.left(LedgerErrorId.UNKNOWN)
```
veado: Error should be more explicit, e.g. `LedgerErrorId.INVALID_PUBKEY` (we don't have such `errorId` yet)

src/main/api/ledger/binance.ts
gromxyz: The problem with this one is type. `getPrefix` accepts different `Network` enum type (the one from the xchainjs package).
But internally we use our `Network` type which is actually a string.

In the end it would be something like: `getPrefix(network === XChainNetwork.Mainnet ? XChainNetwork.Mainnet : XChainNetwork.Testnet)`
Which also doesn't look the best I would say. Do you something better in mind or we can leave it as it is?

veado: We have a helper for that: `toClientNetwork`. It's currently defined in `src/renderer/services/clients/utils.ts`, but needs to be extracted into `src/shared/utils` to make it accesable in `main` and `renderer` processes. Let's do it with a next PR - see #1718 1718

veado (approved):
:+1:

PR #1756, opened 2021-09-13
Move `guard` module + add `is{xyz}Wallet` helpers

veado:
incl. tests


src/shared/utils/guard.test.ts
```diff
@@ -69,6 +78,24 @@ describe('shared/guard', () => {
...
+  describe('isLedgerWallet', () => {
+    it('false for "keystore"', () => {
+      expect(isLedgerWallet('keystore')).toBeFalsy()
+    })
+    it('true for "ledger"', () => {
+      expect(isLedgerWallet('ledger')).toBeTruthy()
+    })
+  })
+
+  describe('isLedgerWallet', () => {
```
gromxyz: Just small typo: `isLedgerWallet` -> `isKeystoreWallet`

gromxyz (dismissed):

src/shared/utils/guard.test.ts
veado: Good catch :smile:

PR #1759, opened 2021-09-13
[Swap] Setting max swap amount missing substracting estimated fee

### LINKED ISSUE #1754: MAX button ends with Insufficient funds
## Prerequisites
Some assets for swapping or sending

## Expected Behavior
The maximum amount of the asset that can be transferred, minus the fees should be filled into the amount field.

## Current Behavior
The total amount of asset in the wallet is filled into the field, without accounting for gas, so if this amount is submitted, the tx will fail due to insufficient funds.

## Steps to Reproduce

https://user-images.githubusercontent.com/45055276/132924003-97ecf8c3-1974-4f76-980a-f5f085bfc1c3.mp4

1. Swap or send some asset
2. Click MAX button for amount
3. Submit transaction
4. Insufficient Funds quickly appears

## Possible Solution
Subtract required gas from total amount in wallet and fill that into the field instead.

## Your Environment - Release Version (skip if not using a release version)
V0.3.14 on Multiple OS

## Discussion at Discord
https://discord.com/channels/838986635756044328/839002220942721024/885209225519828992

gromxyz:
Closes #1754

### DIFF
--- a/src/renderer/components/swap/Swap.tsx
+++ b/src/renderer/components/swap/Swap.tsx
@@ -12,9 +12,8 @@ import {
   formatAssetAmount,
   formatAssetAmountCurrency,
   delay,
-  assetAmount,
-  assetToBase,
-  Chain
+  Chain,
+  AssetRuneNative
 } from '@xchainjs/xchain-util'
 import BigNumber from 'bignumber.js'
 import * as A from 'fp-ts/Array'
@@ -146,7 +145,7 @@ export const Swap = ({
 }: SwapProps) => {
   const intl = useIntl()
 
-  const unlockedWallet = useMemo(() => isLocked(keystore) || !hasImportedKeystore(keystore), [keystore])
+  const unlockedWallet = useMemo(() => !isLocked(keystore) && hasImportedKeystore(keystore), [keystore])
 
   const [targetWalletAddress, setTargetWalletAddress] = useState<O.Option<Address>>(initialTargetWalletAddress)
 
@@ -501,11 +500,12 @@ export const Swap = ({
   // depends on users balances of source asset
   // Decimal always <= 1e8 based
   const maxAmountToSwapMax1e8: BaseAmount = useMemo(() => {
-    // make sure not logged in user can play around with swap
-    if (unlockedWallet) return assetToBase(assetAmount(Number.MAX_SAFE_INTEGER, sourceAssetAmountMax1e8.decimal))
-
+    if (unlockedWallet && sourceAssetAmountMax1e8.gt(swapFees.inFee.amount)) {
+      if (sourceChainAsset === AssetRuneNative) return sourceAssetAmountMax1e8.minus(swapFees.inFee.amount)
+      else return sourceAssetAmountMax1e8.minus(swapFees.inFee.amount.times(2))
+    }
     return sourceAssetAmountMax1e8
-  }, [unlockedWallet, sourceAssetAmountMax1e8])
+  }, [unlockedWallet, sourceAssetAmountMax1e8, swapFees.inFee.amount, sourceChainAsset])
 
   const setAmountToSwapMax1e8 = useCallback(
     (amountToSwap: BaseAmount) => {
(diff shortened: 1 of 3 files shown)


src/renderer/components/swap/Swap.tsx
```diff
@@ -146,7 +145,7 @@ export const Swap = ({
 }: SwapProps) => {
   const intl = useIntl()
 
-  const unlockedWallet = useMemo(() => isLocked(keystore) || !hasImportedKeystore(keystore), [keystore])
+  const unlockedWallet = useMemo(() => !isLocked(keystore) && hasImportedKeystore(keystore), [keystore])
```
veado: Why to reverse the logic and not just rename this function to `lockedWallet`? _I know it's still WIP..._

gromxyz: Yeah, that would actually be good too, will change it back.

src/renderer/components/swap/Swap.utils.test.ts
```diff
@@ -651,4 +652,20 @@ describe('components/swap/utils', () => {
       expect(eqBaseAmount.equals(result, assetToBase(assetAmount(0.01525, inAssetDecimal)))).toBeTruthy()
     })
   })
+
+  describe('maxAmountToSwapMax1e8', () => {
+    it('returns max. balance to swap - estimated fees', () => {
+      const params = {
+        inputBalanceAmount: baseAmount(1000),
```
veado, line 659: Can you add another tests, where `decimal` is `18` (for example for ETH)

src/renderer/components/swap/Swap.tsx
```diff
@@ -501,11 +499,8 @@ export const Swap = ({
   // depends on users balances of source asset
   // Decimal always <= 1e8 based
   const maxAmountToSwapMax1e8: BaseAmount = useMemo(() => {
-    // make sure not logged in user can play around with swap
-    if (unlockedWallet) return assetToBase(assetAmount(Number.MAX_SAFE_INTEGER, sourceAssetAmountMax1e8.decimal))
-
-    return sourceAssetAmountMax1e8
-  }, [unlockedWallet, sourceAssetAmountMax1e8])
+    return Utils.maxAmountToSwapMax1e8(sourceAssetAmountMax1e8, swapFees)
```
veado: Don't remove the case to handle locked / notimported wallet users - they still can play around.

src/renderer/components/swap/Swap.utils.ts
```diff
@@ -257,3 +257,13 @@ export const minBalanceToSwap = (swapFees: Pick<SwapFees, 'inFee'>): BaseAmount
   // Over-estimate balance by 50%
   return feeToCover.times(1.5)
 }
+
+export const maxAmountToSwapMax1e8 = (assetAmount: BaseAmount, swapFees: Pick<SwapFees, 'inFee'>): BaseAmount => {
```
veado: `inFee: BaseAmount` - no need to push down a "complex" object

veado (changes requested):

src/renderer/components/swap/Swap.utils.test.ts
```diff
@@ -651,4 +652,20 @@ describe('components/swap/utils', () => {
...
+        fees: {
+          inFee: {
+            amount: baseAmount(100),
```
veado: Can you add another tests, where `decimal` is `18` (for example for ETH)

src/renderer/components/swap/Swap.utils.ts
veado: Pls change `assetAmount` -> `assetAmountMax1e8` to show it's max. `1e8` based

src/renderer/components/swap/Swap.utils.ts
```diff
@@ -257,3 +257,9 @@ export const minBalanceToSwap = (swapFees: Pick<SwapFees, 'inFee'>): BaseAmount
   // Over-estimate balance by 50%
   return feeToCover.times(1.5)
 }
+
+export const maxAmountToSwapMax1e8 = (assetAmount: BaseAmount, inFeeAmount: BaseAmount): BaseAmount => {
```
veado: `assetAmountMax1e8: BaseAmount, feeAmount: BaseAmount`

src/renderer/components/swap/Swap.utils.ts
```diff
@@ -257,3 +257,9 @@ export const minBalanceToSwap = (swapFees: Pick<SwapFees, 'inFee'>): BaseAmount
...
+  const estimatedFee = to1e8BaseAmount(inFeeAmount.times(1.5))
```
veado: `to1e8BaseAmount` -> `max1e8BaseAmount` (because passed `assetAmount` is `max1e8` based)

veado (changes requested):

src/renderer/components/swap/Swap.utils.ts
veado: Pls add a comment to this function what we expect as params here. Just to remember it in few months from now.

veado (approved):

PR #1763, opened 2021-09-14
Resolve: [AssetInput] Add option to show `i` icon

### LINKED ISSUE #1762: [AssetInput] Add option to show `i` icon
- Right hand side of `max` value
- Whenever a prop `maxInfo` is set, show this icon. By hovering a tooltip will be visible to show text of `maxInfo`

gromxyz:
Closes #1762

### DIFF
--- a/src/renderer/components/uielements/assets/assetInput/AssetInput.tsx
+++ b/src/renderer/components/uielements/assets/assetInput/AssetInput.tsx
@@ -22,6 +22,7 @@ type Props = {
   onFocus?: FP.Lazy<void>
   disabled?: boolean
   className?: string
+  maxInfo?: boolean
 }
 
 /**
@@ -42,6 +43,7 @@ export const AssetInput: React.FC<Props> = (props): JSX.Element => {
     disabled,
     inputProps = {},
     className = '',
+    maxInfo = false,
     onChange,
     onBlur: onBlurHandler = FP.constVoid,
     onFocus: onFocusHandler = FP.constVoid,
@@ -70,6 +72,7 @@ export const AssetInput: React.FC<Props> = (props): JSX.Element => {
           balance={{ amount: maxAmount, asset }}
           onClick={() => onChangeHandler(maxAmount.amount())}
           disabled={disabled}
+          maxInfo={maxInfo}
         />
       </div>
       <div className="asset-input-content" ref={inputRef}>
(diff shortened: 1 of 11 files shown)


gromxyz:
Screenshot 2021-09-14 at 20 15 20

src/renderer/components/uielements/assets/assetInput/AssetInput.tsx
```diff
@@ -22,6 +22,7 @@ type Props = {
   onFocus?: FP.Lazy<void>
   disabled?: boolean
   className?: string
+  maxInfo?: boolean
```
veado: Why not just using a `maxInfoText: string` for props instead of a boolean. With it you can pass any text you want to have for the tooltip. `AssetInput` might not be used for `Swap` only.

src/renderer/components/uielements/button/MaxBalanceButton.styles.ts
```diff
@@ -17,3 +17,16 @@ export const Button = styled(UIButton)`
...
 `
+
+export const InfoLabel = styled.div`
+  width: 18px;
+  height: 18px;
+  margin: 3px;
+  border-radius: 9px;
+  font-size: 11px;
+  text-transform: lowercase;
+  border: 1.5px solid white;
```
veado: What about theming?

src/renderer/components/uielements/button/MaxBalanceButton.tsx
```diff
@@ -39,6 +41,13 @@ export const MaxBalanceButton: React.FC<Props> = (props): JSX.Element => {
           })}
         </Styled.Label>
       </Col>
+      {maxInfo && (
+        <Tooltip
+          overlayStyle={{ maxWidth: '100%', whiteSpace: 'nowrap', fontSize: 11 }}
+          title={intl.formatMessage({ id: 'swap.info.max.fee' })}>
```
veado: Don't use hard-coded text - see first comment about `maxInfoText`

veado (changes requested):

veado (approved):
:+1:

src/renderer/i18n/de/swap.ts
```diff
@@ -7,6 +7,7 @@ const swap: SwapMessages = {
   'swap.input': 'Eingabe',
   'swap.output': 'Ausgabe',
   'swap.recipient': 'Empfänger',
+  'swap.info.max.fee': 'Total asset balance substracted by estimated swap fees - DE',
```
veado, line 10: Guthaben minus voraussichtliche Gebühr

PR #1791, opened 2021-09-26
Prepare release 0.4.1-alpha1

veado:
This is a pre-release to fix withdraw issues we currently have on `chaosnet` only. If this pre-release goes well, another "official" release v0.4.1` will be follow shortly today.


veado:
Close in favour of #1792

PR #183, opened 2020-06-16
Resolve "Add business logic for pricing"

veado:
- [x] Filter out pools using a whitelist of currencies
- [x] Set "price pool" by using filtered pools and selected currency of `HeaderCurrency`
- [x] Calculate prices in `PoolsOverview` using selected `PricePool`

Price selector in header has following logic:
- (1) Provide RUNE and other currencies defined in a whitelist only. Currently USD (TUSDB), BTC, ETH are "whitelisted" (see [current whitelist](https://github.com/thorchain/asgardex-electron/blob/d69706d88a3cc63f67d306da4ee2bef9b45b7aea/src/const.ts#L41)). 
- (2) Provide currencies of (1) only if pools are available for these assets. If none of these pools are available, RUNE can be selected only.
- (3) USD will be selected by default if TUSDB pool is available. If not, RUNE will be selected (similar to (2)
- Last price selection will be stored in localStorage. With next start of application, this selection will be "reloaded", but only if a pool is available for this asset. If not, then go with (3)

Closes #182

PR #1841, opened 2021-10-08
Resolve: [Ledger] Open modal to verify Ledger address

gromxyz:
Closes #183

Screenshot 2021-10-08 at 12 45 26

### DIFF
--- a/src/renderer/components/wallet/settings/WalletSettings.tsx
+++ b/src/renderer/components/wallet/settings/WalletSettings.tsx
@@ -173,9 +173,9 @@ export const WalletSettings: React.FC<Props> = (props): JSX.Element => {
   const [addressToVerify, setAddressToVerify] = useState('')
 
   const renderVerifyAddressModal = useCallback(
-    () => (
+    (ledgerAddress) => (
       <Modal
-        title={'Verify Ledger Address'}
+        title={intl.formatMessage({ id: 'wallet.ledger.verifyAddress.modal.title' })}
         visible={verifyAddressModalVisible}
         onOk={() => {
           setVerifyAddressModalVisible(false)
@@ -191,12 +191,15 @@ export const WalletSettings: React.FC<Props> = (props): JSX.Element => {
         okButtonProps={{ autoFocus: true }}
         cancelText={intl.formatMessage({ id: 'common.cancel' })}>
         <div style={{ textAlign: 'center' }}>
-          Verify address <b>{addressToVerify}</b>
-          <br /> on your device
+          {intl.formatMessage({ id: 'wallet.ledger.verifyAddress.modal.verifyAddress' })}
+          <br />
+          <b>{ledgerAddress}</b>
+          <br />
+          {intl.formatMessage({ id: 'wallet.ledger.verifyAddress.modal.onYourDevice' })}
         </div>
       </Modal>
     ),
-    [addressToVerify, intl, verifyAddressModalVisible]
+    [intl, verifyAddressModalVisible]
   )
 
   const accounts = useMemo(
@@ -205,7 +208,7 @@ export const WalletSettings: React.FC<Props> = (props): JSX.Element => {
         oWalletAccounts,
         O.map((walletAccounts) => (
           <Col key={'accounts'} span={24}>
-            {renderVerifyAddressModal()}
+            {renderVerifyAddressModal(addressToVerify)}
             <Styled.Subtitle>{intl.formatMessage({ id: 'setting.account.management' })}</Styled.Subtitle>
             <Styled.AccountCard>
               <List
@@ -233,7 +236,7 @@ export const WalletSettings: React.FC<Props> = (props): JSX.Element => {
         )),
         O.getOrElse(() => <></>)
       ),
-    [oWalletAccounts, renderVerifyAddressModal, intl, renderAddress]
+    [oWalletAccounts, renderVerifyAddressModal, addressToVerify, intl, renderAddress]
   )
 
   return (
--- a/src/renderer/components/wallet/settings/WalletSettings.tsx
+++ b/src/renderer/components/wallet/settings/WalletSettings.tsx
@@ -20,6 +20,7 @@ import { getChainAsset, isBnbChain } from '../../../helpers/chainHelper'
 import { ValidatePasswordHandler, WalletAccounts, WalletAddress } from '../../../services/wallet/types'
 import { walletTypeToI18n } from '../../../services/wallet/util'
 import { InfoIcon } from '../../uielements/info'
+import { Modal } from '../../uielements/modal'
 import * as Styled from './WalletSettings.styles'
 
 type Props = {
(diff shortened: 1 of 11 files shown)


src/renderer/i18n/de/wallet.ts
```diff
@@ -83,7 +83,10 @@ const wallet: WalletMessages = {
   'wallet.upgrade.feeError': 'Die Upgrade Gebühr {fee} ist nicht über Dein Guthaben {balance} gedeckt',
   'wallet.validations.lessThen': 'Der eingegebene Betrag sollte weniger als {value} betragen',
   'wallet.validations.graterThen': 'Der eingegebene Betrag sollte höher als {amount} betragen',
-  'wallet.validations.shouldNotBeEmpty': 'Der eingegebene Werte sollte nicht leer sein.'
+  'wallet.validations.shouldNotBeEmpty': 'Der eingegebene Werte sollte nicht leer sein.',
+  'wallet.ledger.verifyAddress.modal.title': 'Verify Ledger Address',
+  'wallet.ledger.verifyAddress.modal.verifyAddress': 'Verify address',
+  'wallet.ledger.verifyAddress.modal.onYourDevice': 'on your device'
```
veado:
```
  'wallet.ledger.verifyAddress.modal.description': 'Überprüfe die Addresse {address} auf Deinem Ledger',
 ```

File: src/renderer/i18n/types.ts [hunk 2]
```diff
@@ -199,6 +199,9 @@ type WalletMessageKey =
   | 'wallet.validations.lessThen'
   | 'wallet.validations.graterThen'
   | 'wallet.validations.shouldNotBeEmpty'
+  | 'wallet.ledger.verifyAddress.modal.title'
+  | 'wallet.ledger.verifyAddress.modal.verifyAddress'
+  | 'wallet.ledger.verifyAddress.modal.onYourDevice'
```
veado: `'wallet.ledger.verifyAddress.modal.description'`

src/renderer/i18n/en/wallet.ts
```diff
@@ -80,7 +80,10 @@ const wallet: WalletMessages = {
   'wallet.upgrade.feeError': 'Upgrade fee {fee} is not covered by your balance {balance}',
   'wallet.validations.lessThen': 'Should be less then {value}',
   'wallet.validations.graterThen': 'Should be grater then {value}',
-  'wallet.validations.shouldNotBeEmpty': 'Should not be empty'
+  'wallet.validations.shouldNotBeEmpty': 'Should not be empty',
+  'wallet.ledger.verifyAddress.modal.title': 'Verify Ledger Address',
+  'wallet.ledger.verifyAddress.modal.verifyAddress': 'Verify address',
+  'wallet.ledger.verifyAddress.modal.onYourDevice': 'on your device'
```
veado: You can use a placeholder for `address`
```
  'wallet.ledger.verifyAddress.modal.description': 'Verify address {address} on your device',
 ```

File: src/renderer/components/wallet/settings/WalletSettings.tsx [hunk 4]
```diff
@@ -162,12 +169,46 @@ export const WalletSettings: React.FC<Props> = (props): JSX.Element => {
     ]
   )
 
+  const [verifyAddressModalVisible, setVerifyAddressModalVisible] = useState(false)
+  const [addressToVerify, setAddressToVerify] = useState('')
```
veado: You can check `addressToVerify` to open modal or not, don't you? No need for another state `verifyAddressModalVisible` then.

src/renderer/components/wallet/settings/WalletSettings.tsx
```diff
@@ -162,12 +169,46 @@ export const WalletSettings: React.FC<Props> = (props): JSX.Element => {
...
+      <Modal
+        title={intl.formatMessage({ id: 'wallet.ledger.verifyAddress.modal.title' })}
+        visible={verifyAddressModalVisible}
+        onOk={() => {
+          setVerifyAddressModalVisible(false)
+          setAddressToVerify('')
+        }}
+        onCancel={() => {
+          setVerifyAddressModalVisible(false)
+          setAddressToVerify('')
```
veado: Is there a way to remove Ledger from ASGARDEX in case of `cancel`? Re-name `cancel` to `reject` then.

veado (changes requested):

src/renderer/i18n/en/wallet.ts
gromxyz: I've tried using placeholder, but there is also`<b></b>` inside the string so it treats them as placeholder too and then not renders correctly.
So that's the reason why I separated these two translations

src/renderer/components/wallet/settings/WalletSettings.tsx
gromxyz: Actually, yes 😅

src/renderer/components/wallet/settings/WalletSettings.tsx
gromxyz: Will be added 👍

src/renderer/i18n/en/wallet.ts
veado: You will find an example in our code how to do add html tags (search for `FormattedMessage`).

src/renderer/components/wallet/settings/WalletSettings.tsx
```diff
@@ -169,37 +172,46 @@ export const WalletSettings: React.FC<Props> = (props): JSX.Element => {
...
-  const [verifyAddressModalVisible, setVerifyAddressModalVisible] = useState(false)
-  const [addressToVerify, setAddressToVerify] = useState('')
-
+  const DEFAULT_ADDRESS_TO_VERIFY = useMemo(() => {
+    return {
+      address: '',
+      chain: THORChain
+    }
+  }, [])
+  const [addressToVerify, setAddressToVerify] = useState(DEFAULT_ADDRESS_TO_VERIFY)
```
veado: `useState<O.Option<Address>>` and you don't need an empty `DEFAULT_ADDRESS_TO_VERIFY` (which might be confusing)

src/renderer/i18n/de/common.ts
```diff
@@ -24,6 +24,7 @@ const common: CommonMessages = {
   'common.submit': 'Abschicken',
   'common.confirm': 'Bestätigen',
   'common.cancel': 'Abbrechen',
+  'common.reject': 'Reject - DE',
```
veado: `'common.reject': Ablehnen`

veado (changes requested):

src/renderer/components/wallet/settings/WalletSettings.tsx
gromxyz: The problem is that I have the chain also saved here.
Is there a way to get the chain based on address, so I don't have to save the chain here?

veado: :eye: https://github.com/thorchain/asgardex-electron/pull/1841/commits/6324ae2a783632a420a2d43dd1f8d86704108370

veado (approved):

Issue #1811, opened 2021-10-01
[History] Update layout 


veado:
- Remove header from table
- Add `AccountAddressSelector`
- Add link to `viewblock` (depending on selected address + history type)

PR #1842, opened 2021-10-09
Introduce useMidgardHistoryActions hook

veado:
to keep state handling in `view` instead of `service` level. It keeps `PoolHistoryView` + `PoolActionsHistoryView` more clean by removing (duplicated) logic from there. No need of calling `resetActionsData` anymore.

- [x] `useMidgardHistoryActions` hook
- [x] Remove state handling from `services/midgard&poolHistoryActions`
- [x] Some renaming: `poolActionXYZ` -> `actionXYZ`, `actionsXYZ` -> `historyXYZ`, `view/wallet/PoolActionsHistory` ->
`view/wallet/history`

Part of #1811


src/renderer/services/midgard/actions.ts
```diff
@@ -0,0 +1,62 @@
+import * as RD from '@devexperts/remote-data-ts'
```
veado, line 1: Note: Almost no change in this file. It has just been renamed: `poolActionsHistory` -> `actions`. Also it's state handling is in `useMidgardHistory` hook now

gromxyz (approved):

PR #1885, opened 2021-10-26
Resolve: [Header] Tooltip to explain VOLUME (24h)

gromxyz:
Screenshot 2021-10-26 at 10 18 06

Closes #1874

### DIFF
--- a/src/renderer/i18n/de/common.ts
+++ b/src/renderer/i18n/de/common.ts
@@ -86,6 +86,7 @@ const common: CommonMessages = {
   'common.theme.light': 'Tag modus',
   'common.theme.dark': 'Nacht modus',
   'common.volume24': 'Volumen (24h)',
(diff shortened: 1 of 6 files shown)


src/renderer/i18n/de/common.ts
```diff
@@ -86,6 +86,7 @@ const common: CommonMessages = {
   'common.theme.light': 'Tag modus',
   'common.theme.dark': 'Nacht modus',
   'common.volume24': 'Volumen (24h)',
+  'common.volume24.description': '24h Handlesvolumen von Swaps, Pooleinzahlungen und abhebungen',
```
veado: Typo: `-abhebungen` (Note missing `-`)

gromxyz: I thought it was typo from your side

gromxyz: Fixed 👍

veado (approved):

PR #1892, opened 2021-10-27
[Swap] Introduce LedgerConfirmationModal 

veado:
- [x] `LedgerConfirmationModal` incl. story
- [x] Use it ^ in `Swap`
- [x] Update `i18n`
- [x] Update `Ledger` icon (we don't use it here - but maybe later)
- [x] Few tweaks for storybook (configuration, packages, other stories)

Screenshot from 2021-10-27 16-33-24

Close #1887


gromxyz (approved):
👍

PR #1931, opened 2021-11-04
Resolve: [Swap] Update slippage tolerance

### LINKED ISSUE #1929: [Swap] Update slippage tolerance
- Add an `i` icon with following explanation:

 ```
    # EN
    The higher the percentage, the more slippage you will accept. More slippage includes also a wider range for covering estimated fees to avoid aborted swaps.
    # DE
    Je höher die Prozentangabe, je höher akzeptierst Du ein Slippage. Mehr Slippage bedeutet zugleich ein größerer Spielraum zur Abdeckung der geschätzten Gebühren, um fehlgeschlagene Swaps zu vermeiden. 
```

 - Update calculation to add `999` (index for ASGARDEX) to price limits to decrease (but not increase) values. Decrease values includes a (minimal) chance to let swaps fail.
     - `1000000` -> BEFORE: `1000999` / NOW: `9999999`
     - `1002111` -> BEFORE: `1002999` / NOW: `1001999`
     - `1009888` -> BEFORE: `1009999` / NOW: `1008999`
- Add tests for ^

### PR DESCRIPTION
**Pull request by gromxyz**
Screenshot 2021-11-04 at 17 52 43

Closes #1929

### DIFF
--- a/src/renderer/components/swap/Swap.utils.ts
+++ b/src/renderer/components/swap/Swap.utils.ts
@@ -157,7 +157,7 @@ export const getSwapData = ({
 export const getSwapLimit = (swapResultAmountMax1e8: BaseAmount, slipTolerance: SlipTolerance): BaseAmount => {
   const swapLimit: BaseAmount = swapResultAmountMax1e8.times(1.0 - slipTolerance * 0.01)
   const swapLimitWithIdentifier = +swapLimit.amount().toString().slice(0, -3).concat(ASGARDEX_SWAP_IDENTIFIER) - 1000
-  return baseAmount(bn(swapLimitWithIdentifier))
+  return baseAmount(bn(swapLimitWithIdentifier <= 999 ? 999 : swapLimitWithIdentifier))
 }
 
 export const pickPoolAsset = (assets: PoolAssetDetails, asset: Asset): O.Option<PoolAssetDetail> =>
--- a/src/renderer/components/swap/Swap.utils.ts
+++ b/src/renderer/components/swap/Swap.utils.ts
@@ -156,8 +156,11 @@ export const getSwapData = ({
  */
 export const getSwapLimit = (swapResultAmountMax1e8: BaseAmount, slipTolerance: SlipTolerance): BaseAmount => {
   const swapLimit: BaseAmount = swapResultAmountMax1e8.times(1.0 - slipTolerance * 0.01)
-  const swapLimitWithIdentifier = +swapLimit.amount().toString().slice(0, -3).concat(ASGARDEX_SWAP_IDENTIFIER) - 1000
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/renderer/components/swap/Swap.utils.ts [hunk 1]
```diff
@@ -156,7 +156,7 @@ export const getSwapData = ({
  */
 export const getSwapLimit = (swapResultAmountMax1e8: BaseAmount, slipTolerance: SlipTolerance): BaseAmount => {
   const swapLimit: BaseAmount = swapResultAmountMax1e8.times(1.0 - slipTolerance * 0.01)
-  const swapLimitWithIdentifier: number = +swapLimit.amount().toString().slice(0, -3).concat(ASGARDEX_SWAP_IDENTIFIER)
+  const swapLimitWithIdentifier = +swapLimit.amount().toString().slice(0, -3).concat(ASGARDEX_SWAP_IDENTIFIER) - 1000
```
veado: Pls add tests for lower `swapLimit` (< `999` , e.g. `100`, `200` etc.) to see `swapLimitWithIdentifier` is still `999`

gromxyz: Those are really rare cases and if the user deals with those values, then setting limit would probably be incorrect in this case.

gromxyz: But I've updated the code in order to handle these cases.

src/renderer/components/swap/Swap.utils.ts
```diff
@@ -156,8 +156,8 @@ export const getSwapData = ({
  */
 export const getSwapLimit = (swapResultAmountMax1e8: BaseAmount, slipTolerance: SlipTolerance): BaseAmount => {
   const swapLimit: BaseAmount = swapResultAmountMax1e8.times(1.0 - slipTolerance * 0.01)
-  const swapLimitWithIdentifier: number = +swapLimit.amount().toString().slice(0, -3).concat(ASGARDEX_SWAP_IDENTIFIER)
-  return baseAmount(bn(swapLimitWithIdentifier))
+  const swapLimitWithIdentifier = +swapLimit.amount().toString().slice(0, -3).concat(ASGARDEX_SWAP_IDENTIFIER) - 1000
+  return baseAmount(bn(swapLimitWithIdentifier <= 999 ? 999 : swapLimitWithIdentifier))
```
veado: Use `ASGARDEX_SWAP_IDENTIFIER` instead of `999`. Change it before to be a number in `renderer/const.ts`

veado (changes requested):

veado (commented):
Update position of `i` icon - it is related to `slippage tolerance`, not to slippage.

veado (approved):

Issue #1987, opened 2021-12-17
ADD: Stagenet


thorchain-admin:
- [x] Stagenet wallet variable
- [x] Stagenet Midgard baseURL `stagenet.midgard.ninerealms.capital`
- [x] When stagenet, init `xchain-thorchain` into `stagenet`, load all others into `mainnet`
- [x] Disable THORChain ledger when in stagenet (since does not support)

veado:
Resolved by #2000

PR #2000, opened 2021-12-29
Issue 1987/add stagenet

polaris-9r:
### Description
This ticket enables a user to switch to stagenet for THORChain, but leaves all other chains as chaosnet. This PR includes the following major changes

- Updating `@xchainjs` libraries to new breaking change versions that add stagenet functionality for the base client class
- Adding front-end selector functionality to select a `Stagenet` network, with coloring in red
- Disable adding a ledger for a stagenet TC account, as it has not been enabled
- Adding testing to validate stagenet network settings, plus new testing for `isEnabledLedger` check.

### Notes
I was able to verify that we were able to create an `sthor` prefixed wallet addresses via a dev version of the electron app, and transfer `sthor` RUNE into that account.

Screenshots of new FE styling:

Chaosnet
image
image

Stagenet
image
image

~NOTE: Leaving this in draft form at the moment since this is dependent on landing a change in xchainjs to add stagenet environment config: https://github.com/xchainjs/xchainjs-lib/pull/456~

~TODO:~
- ~Land xchainjs changes to add stagenet~
- ~Uncomment changes that reference Network.Stagenet~
- ~Run through CI/CD to see if there are any other issues~
- Show pool information for mainnet for all non-TC assets, only stagenet for TC

#1987


veado (commented):
DRAFT LGTM. 

Just two (small) things so far:
(1) Pls update helper [toClientNetwork](https://github.com/thorchain/asgardex-electron/blob/e034b25071e0277fe7f6c4ebeb45daf0620010b0/src/shared/utils/client.ts#L5-L6)
(2) In case you want to change color of network label in header, here is the place to do https://github.com/thorchain/asgardex-electron/blob/eb574a58f4d6ebc2a930be075b3ab5cdccc840c0/src/renderer/components/header/HeaderComponent.styles.ts#L85-L94

polaris-9r:
@veado Wanted your perspective on one last part of this PR. So as you can see from the screenshots in my first post, the pool values are different between chaosnet and stagenet because they are hitting the mainnet vs stagenet versions of midgard. I can confirm though that in the wallet tab that the non-TC chain wallet addresses are the same values. Question is:

- Do we want the pools tab to reflect all mainnet pool values? All stagenet pool values? (i.e. simply setting the env. var for `REACT_APP_MIDGARD_STAGENET_URL`)
- Do we want the pools tab to show some hybrid of stagenet for TC and mainnet for non-TC?

src/renderer/const.ts
```diff
@@ -235,11 +239,13 @@ export const ASGARDEX_SWAP_IDENTIFIER = 999
 
 export const RECOVERY_TOOL_URL: Record<Network, string> = {
   testnet: 'https://testnet.thorswap.finance/pending',
+  stagenet: 'https://stagenet.thorswap.finance/pending',
```
polaris-9r, line 242: @veado Wanted confirmation as to whether this should be stagenet specific or the mainnet value.

src/renderer/const.ts
```diff
@@ -235,11 +239,13 @@ export const ASGARDEX_SWAP_IDENTIFIER = 999
...
   mainnet: 'https://app.thorswap.finance/pending'
 }
 
 export const ASYM_DEPOSIT_TOOL_URL: Record<Network, string> = {
   testnet: 'https://testnet.thorswap.finance/',
+  stagenet: 'https://stagenet.thorswap.finance/',
```
polaris-9r, line 248: @veado See above

.env.sample
```diff
@@ -5,11 +5,14 @@ REACT_APP_CHAINS_ENABLED=THOR,BNB,BTC,LTC,BCH,ETH
 
 # Midgard
 REACT_APP_MIDGARD_TESTNET_URL=https://testnet.midgard.thorchain.info
+REACT_APP_MIDGARD_STAGENET_URL=XXX
```
pluto-x1: https://stagenet-midgard.ninerealms.com

.env.sample
```diff
@@ -5,11 +5,14 @@ REACT_APP_CHAINS_ENABLED=THOR,BNB,BTC,LTC,BCH,ETH
...
 REACT_APP_MIDGARD_MAINNET_URL=https://midgard.thorchain.info
 
 # THORnode
 REACT_APP_TESTNET_THORNODE_API=https://testnet.thornode.thorchain.info
+REACT_APP_STAGENET_THORNODE_API=XXX
```
pluto-x1: https://stagenet-thornode.ninerealms.com

.env.sample
```diff
@@ -5,11 +5,14 @@ REACT_APP_CHAINS_ENABLED=THOR,BNB,BTC,LTC,BCH,ETH
...
 REACT_APP_MAINNET_THORNODE_API=https://thornode.thorchain.info
+REACT_APP_STAGENET_THORNODE_RPC=XXX
```
pluto-x1: https://stagenet-rpc.ninerealms.com

src/main/api/url.ts
```diff
@@ -24,12 +24,15 @@ const EXTERNALS_WHITELIST = [
...
+  'app.thorswap.finance',
   'viewblock.io',
   'midgard.thorchain.info',
   'testnet.midgard.thorchain.info',
+  'stagenet-midgard.ninerealms.com',
   'thornode.thorchain.info',
-  'testnet.thornode.thorchain.info'
+  'testnet.thornode.thorchain.info',
+  'stagenet-thornode.ninerealms.com'
 ]
```
pluto-x1, line 37: ADD: stagenet-rpc.ninerealms.com

pluto-x1 (commented):
nice, lgtm

.env changes are optional

veado:
Wallet -> Assets -> ETH
> UNSUPPORTED GETDEFAULTPROVIDER NETWORK (OPERATION="GETDEFAULTPROVIDER", NETWORK="STAGENET", CODE=NETWORK_ERROR, VERSION=PROVIDERS/5.4.5)

Screenshot from 2022-01-04 12-25-58

`network` should be `mainnet`(not `stagenet`) It looks like an issue at latest `xchain-ethereum`.

veado:
Wallet -> Assets -> BTC
> xhr.js:210 GET https://api.haskoin.com/haskoin-store/btc/address/{address}/balance net::ERR_FAILED 404

Same for chaosnet, It might be an issue with haskoin api. Can we use `https://haskoin.ninerealms.com/` ?


Screenshot from 2022-01-04 12-30-14

veado:
> @veado Wanted your perspective on one last part of this PR. So as you can see from the screenshots in my first post, the pool values are different between chaosnet and stagenet because they are hitting the mainnet vs stagenet versions of midgard. I can confirm though that in the wallet tab that the non-TC chain wallet addresses are the same values. Question is:
> 
>     - Do we want the pools tab to reflect all mainnet pool values? All stagenet pool values? (i.e. simply setting the env. var for `REACT_APP_MIDGARD_STAGENET_URL`)
> 
>     - Do we want the pools tab to show some hybrid of stagenet for TC and mainnet for non-TC?

I would keep it separate (stagenet -> stagenet pools only, mainnet -> mainnet pools only - but no hybrid) to handle differences more easily. Since we are using different RUNE (`thor` vs. `sthor`) addresses, pairing RUNE will be different. Not sure if this might be an issue by mixing stagenet/mainnet pools.

src/renderer/const.ts
veado: It seems https://stagenet.thorswap.finance/pending is accessible, so let's use it.

src/renderer/const.ts
veado: https://github.com/thorchain/asgardex-electron/pull/2000#discussion_r778048303

veado:
It seems `viewblock` does not support `stagenet` yet, that's why external links for any tx are going to https://viewblock.io/thorchain instead of https://viewblock.io/thorchain/tx/{hash}?network=stagenet (which is not accessible atm)

Example: `pools` -> `detail`

Screenshot from 2022-01-04 18-18-27

Three options:
- Ignore this issue and just let external tx links go to `https://viewblock.io/thorchain`
- Hide all external icons + view transaction links in case of `stagenet`
- Fix tx urls to be `https://viewblock.io/thorchain/tx/{hash}?network=stagenet` (in hope it will be such url)

polaris-9r:
> It seems `viewblock` does not support `stagenet` yet, that's why external links for any tx are going to https://viewblock.io/thorchain instead of [https://viewblock.io/thorchain/tx/{hash}?network=stagenet](https://viewblock.io/thorchain/tx/%7Bhash%7D?network=stagenet) (which is not accessible atm)
> 
> Example: `pools` -> `detail`
> 
> Screenshot from 2022-01-04 18-18-27
> 
> Three options:
> 
>     * Ignore this issue and just let external tx links go to `https://viewblock.io/thorchain`
> 
>     * Hide all external icons + view transaction links in case of `stagenet`
> 
>     * Fix tx urls to be `https://viewblock.io/thorchain/tx/{hash}?network=stagenet` (in hope it will be such url)

Yeah, given this lack of functionality, I think I'd personally want to err on the side of not showing transactions at all in case of stagenet, in line with what we were saying that stagenet should only see stagenet, mainnet should only see mainnet, etc. I'll put some gatekeeping for now and we can figure out when viewblock is going to support stagenet.

veado (approved):
300fce1 fixes xchain-* related issues. Agree, urls for viewblock are not a blocker - the viewblock team will provide stagenet soon (see Discord message https://discord.com/channels/838986635756044328/846612413477945384/928043598367326228)

Happy to merge now, thanks again for your great work @polaris-9r!

PR #2141, opened 2022-03-15
Feature/update fr translation

Weyland2093:
Feature/update fr translation


veado:
Thx @Weyland2093 - updates will go into next release

PR #2170, opened 2022-03-26
Add more slippage options

WojciechKo:


WojciechKo:
I was surprised that default slippage is set as high as 5%.
What do you think about setting it to 1% like it is ThorSwap? Seems it work just fine.

veado:
> I was surprised that default slippage is set as high as 5%. What do you think about setting it to 1% like it is ThorSwap? 

~Sounds good. Pls update it in `DEFAULT_SLIP_TOLERANCE` + add to your PR.~

^ Done #2175

PR #2403, opened 2022-09-29
Feature/update fr translation

Weyland2093:
updated translation for french version !


veado:
Thanks @Weyland2093 !

PR #2471, opened 2022-12-03
Translation ru3 dec

TreefeedXavier:
common.ts and pools.ts translation


veado:
Thanks @TreefeedXavier!

PR #2496, opened 2022-12-20
Translation ru20 dec (pt2)

TreefeedXavier:
RU translation for savers


veado:
Thx @TreefeedXavier !

PR #2518, opened 2023-01-03
- RU translation for deposit.ts

TreefeedXavier:
Recent RU translation for 1.18.6


veado:
Thank you @TreefeedXavier

PR #2531, opened 2023-02-01
- RU translation for deposit.ts

TreefeedXavier:
Recent for 01 Feb 2023


veado:
Thanks @TreefeedXavier!

PR #2546, opened 2023-07-18
adding support for synths, view in wallets, swapping

### LINKED ISSUE #2545: Add: Support for synths
## Description

- [ ]  View synths in wallet 
- [ ]  View history transactions fro synths
- [ ] Support swapping of Synths

Thorian1te:


src/renderer/services/chain/fees/common.ts
```diff
@@ -17,7 +17,7 @@ const {
  */
 export const poolOutboundFee$ = (asset: Asset): PoolFeeLD => {
   // special case for RUNE - not provided in `inbound_addresses` endpoint
-  if (isRuneNativeAsset(asset)) {
+  if (isRuneNativeAsset(asset) || asset.synth) {
     return FP.pipe(
       THOR.fees$(),
       liveData.map((fees) => ({ amount: fees.fast.times(3), asset: AssetRuneNative }))
```
GrassRootsCrypto, line 23: Why 3 x fast here? I think it is a flat 0.02 fee

src/shared/utils/asset.ts
```diff
@@ -12,10 +12,18 @@ import {
   AssetRuneERC20,
   AssetRuneERC20Testnet
 } from '@xchainjs/xchain-thorchain'
+import { assetFromStringEx } from '@xchainjs/xchain-util'
+
+const AssetSynthBtc = assetFromStringEx('BTC/BTC')
```
GrassRootsCrypto, line 17: Is there a reason why other synth assets are not added? E,g, ETH/ETH?
I see in render/helpers/assetHelpers isBtcAssetSynth is added for synth asset comparison but there is nothing for ETH and other L1 assets.

Thorian1te: at this stage I don't think they are needed that much

src/renderer/services/chain/fees/common.ts
Thorian1te: it is not used

